PR #1, opened 2016-07-26
Deal with Bad Migration When Extending Persisted Entity

montanehamilton:
Added a test to demonstrate the behavior, have to jump through some hoops to get there but that might be useful in the future.

Fix uses the IsOfType as the type if the type is not specified and there is not a prior match using the old logic. That is not enough as the matching is more eager than before so also keep track of matches made and don't reuse a target fragment that has already been matched.

### DIFF
--- a/src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
+++ b/src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
@@ -84,7 +84,7 @@ namespace System.Data.Entity.Migrations.Infrastructure
                 {
                     EdmItemCollection = storageMappingItemCollection.EdmItemCollection,
                     StoreItemCollection = storageMappingItemCollection.StoreItemCollection,
-                    StoreEntityContainer 
+                    StoreEntityContainer
                         = storageMappingItemCollection.StoreItemCollection.GetItems<EntityContainer>().Single(),
                     EntityContainerMapping
                         = storageMappingItemCollection.GetItems<EntityContainerMapping>().Single(),
@@ -109,11 +109,11 @@ namespace System.Data.Entity.Migrations.Infrastructure
                 };
 
             return Diff(
-                source, 
-                target, 
-                modificationCommandTreeGenerator, 
-                migrationSqlGenerator, 
-                sourceModelVersion, 
+                source,
+                target,
+                modificationCommandTreeGenerator,
+                migrationSqlGenerator,
+                sourceModelVersion,
                 targetModelVersion);
         }
 
@@ -328,6 +328,27 @@ namespace System.Data.Entity.Migrations.Infrastructure
                     / (entityType1.DeclaredMembers.Count + entityType2.DeclaredMembers.Count)) > 0.80;
         }
 
+        private bool SourceAndTargetMatch(EntityType sourceEntityType, EntityTypeMapping sourceEntityTypeMapping, EntityType targetEntityType, EntityTypeMapping targetEntityTypeMapping)
+        {
+            return (sourceEntityTypeMapping.EntityType != null
+                    && targetEntityTypeMapping.EntityType != null
+                    && sourceEntityType == sourceEntityTypeMapping.EntityType
+                    && targetEntityType == targetEntityTypeMapping.EntityType)
+                   || (sourceEntityTypeMapping.EntityType == null
+                       && targetEntityTypeMapping.EntityType == null
+                       && sourceEntityTypeMapping.IsOfTypes.Contains(sourceEntityType)
+                       && targetEntityTypeMapping.IsOfTypes.Contains(targetEntityType)
+                       && sourceEntityTypeMapping.IsOfTypes.Except(new[] { sourceEntityType }).Select(et => et.Name)
+                           .SequenceEqual(targetEntityTypeMapping.IsOfTypes.Except(new[] { targetEntityType }).Select(et => et.Name)));
+        }
+
+        private bool MappingTypesAreIdentical(EntityTypeMapping sourceEntityTypeMapping, EntityTypeMapping targetEntityTypeMapping)
+        {
+            var canonicalSourceEntityMappingType = sourceEntityTypeMapping.EntityType ?? sourceEntityTypeMapping.IsOfTypes.First();
+            var canonicalTargetEntityMappingType = targetEntityTypeMapping.EntityType ?? targetEntityTypeMapping.IsOfTypes.First();
+            return canonicalSourceEntityMappingType.FullName == canonicalTargetEntityMappingType.FullName;
+        }
+
         [SuppressMessage("Microsoft.Maintainability", "CA1502:AvoidExcessiveComplexity")]
         private IEnumerable<Tuple<MappingFragment, MappingFragment>> FindMappingFragmentPairs(
             ICollection<Tuple<EntityType, EntityType>> entityTypePairs)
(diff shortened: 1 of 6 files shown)


src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
```diff
@@ -328,6 +328,27 @@ where m1.EdmEquals(m2)
...
 
+        private bool SourceAndTargetMatch(EntityType sourceEntityType, EntityTypeMapping sourceEntityTypeMapping, EntityType targetEntityType, EntityTypeMapping targetEntityTypeMapping)
+        {
+            return (sourceEntityTypeMapping.EntityType != null
+                    && targetEntityTypeMapping.EntityType != null
+                    && sourceEntityType == sourceEntityTypeMapping.EntityType
+                    && targetEntityType == targetEntityTypeMapping.EntityType)
+                   || (sourceEntityTypeMapping.EntityType == null
+                       && targetEntityTypeMapping.EntityType == null
+                       && sourceEntityTypeMapping.IsOfTypes.Contains(sourceEntityType)
```
IDisposable: How costly is `IsOfTypes`? Would it be better to unroll this horrifically complex boolean so we can read it and assign the `IsOfTypes` collection to a local variable?

IDisposable: Follow up... looks like it's doing this for every call :(
`return new ReadOnlyCollection<EntityTypeBase>(new List<EntityTypeBase>(m_isOfEntityTypes.Values));`

So that's bad.

montanehamilton: This bit of code I did not want to redo, we tried multiple versions where it would break other tests. I opted to leave it alone, albeit sticking it in a function rather than attempt to unroll it.

Honestly this entire decision should be a separate class with some descriptive tests to understand the different situations it should take into account. Given that does not exist I opted to leave that code alone.

I'd be happy to improve it with some requirements, but with the hours invested in playing with it I opted to leave it alone.

IDisposable: How's this?

```
        private bool SourceAndTargetMatch(EntityType sourceEntityType, EntityTypeMapping sourceEntityTypeMapping, EntityType targetEntityType, EntityTypeMapping targetEntityTypeMapping)
        {
            if (sourceEntityTypeMapping.EntityType != null
                && targetEntityTypeMapping.EntityType != null)
            {
                if (sourceEntityType == sourceEntityTypeMapping.EntityType
                    && targetEntityType == targetEntityTypeMapping.EntityType)
                {
                    return true;
                }
            }
            else
            {
                var sourceTypes = sourceEntityTypeMapping.IsOfTypes;

                if (sourceTypes.Contains(sourceEntityType))
                {
                    var targetTypes = targetEntityTypeMapping.IsOfTypes;

                    if (targetTypes.Contains(targetEntityType))
                    {
                        var sourceTypeNames = sourceTypes.Except(new[] { sourceEntityType }).Select(et => et.Name);
                        var targetTypeNames = targetTypes.Except(new[] { targetEntityType }).Select(et => et.Name);

                        if (sourceTypeNames.SequenceEqual(targetTypeNames))
                        {
                            return true;
                        }
                    }
                }
            }

            return false;
        }
```

montanehamilton:
Added IDisposable's recommended refactor as it passed all tests as well.

test/EntityFramework/UnitTests/Migrations/Infrastructure/EdmModelDifferTests.cs
```diff
@@ -3650,6 +3654,93 @@ var operations
...
+
+        [Table("PlatformUser")]
+        public class PlatformUser : IPlatformEntity
+        {
+            public long Id { get; set; }
+
+            public string UserName { get; set; }
+        }
+
+        public class TypeGenerator
```
anpete: Why do we need this?

src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
```diff
@@ -328,6 +328,48 @@ where m1.EdmEquals(m2)
                     / (entityType1.DeclaredMembers.Count + entityType2.DeclaredMembers.Count)) > 0.80;
         }
 
+        private bool SourceAndTargetMatch(EntityType sourceEntityType, EntityTypeMapping sourceEntityTypeMapping, EntityType targetEntityType, EntityTypeMapping targetEntityTypeMapping)
```
anpete: Can this method be static?

src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
```diff
@@ -328,6 +328,48 @@ where m1.EdmEquals(m2)
...
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            return false;
+        }
+
+        private bool MappingTypesAreIdentical(EntityTypeMapping sourceEntityTypeMapping, EntityTypeMapping targetEntityTypeMapping)
```
anpete: Can this method be static?

test/EntityFramework/UnitTests/Migrations/Infrastructure/EdmModelDifferTests.cs
montanehamilton: If the extension to the model is compiled the model generator finds it and builds the initial model with the extension already in it. I needed the initial model to be created without the extension, then add the extension as part of a migration, then perform the comparison between two models. This matches the scenario of deliver a database, then update that database with a new model later. The scenario we are getting the exception while doing now.

src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
montanehamilton: Yes, will commit as static.

src/EntityFramework/Migrations/Infrastructure/EdmModelDiffer.cs
montanehamilton: Yes, will commit as static.

test/EntityFramework/UnitTests/Migrations/Infrastructure/EdmModelDifferTests.cs
anpete: Can you just .Ignore the derived type when building the first model?

montanehamilton: That certainly seems simpler, let me give that a go.

montanehamilton: The ignore gives me the same result, commit coming.

montanehamilton: Removed the code generation and all NuGets added to support it.

montanehamilton:
@anpete , I am not seeing any nightly builds yet, is it me or are they not working yet?

natemcmaster:
@montanehamilton your changes should be in this nightly build:  http://myget.org/feed/aspnetwebstacknightly/package/nuget/EntityFramework/6.2.0-alpha1-50813

montanehamilton:
Thank you, I think there was some user error on my part since the build was there on the 15th!

PR #5, opened 2016-07-29
Issue warning when EF Core commands are also installed in solution

natemcmaster:
Matching feature of EF Core side https://github.com/aspnet/EntityFramework/pull/5565

cc @bricelam @rowanmiller


bricelam:
:shipit:

PR #36, opened 2016-08-05
Implement DbFunction for Like operator.

brandondahler:
Implementation for CodePlex Issues [#2388](https://entityframework.codeplex.com/workitem/2388) and [#1323](https://entityframework.codeplex.com/workitem/1323). 

It implements it by 
1. Providing an extension method/static method on DbFunctions and a static method in EntityFunctions for backwards compatibility. 
2. Providing a new CallTranslator (LikeTranslator) to handle these functions being called within expressions. 
3. Providing a TranslateLike helper method that turns the static method invocation into a LikeExpression 
4. Providing a means to turn a character parameter into a single-character string. Doing this allows Like to have both Like(string, string, char) and Like(string, string, string) signatures that are exactly equivalent. 
5. Handling the case where a provider specifies that it does not accept a like operator escape character by throwing an ProviderIncompatibleException with a newly created message. I found it important to have both Like(string, string, char) and Like(string, string, string) because char allows us to enforce the single-character string escape requirement when specifying a literal, while string gives more freedom to your escapeCharacter expression to use a dynamic escape character. 

Tests have been added to match similar tests of DbFunctions.AsUnicode(...). 

Note: I am not a part of the EF team, I cannot provide information about when/if this pull request will be accepted nor when/if it would be actually released. 

This request was migrated from CodePlex Contribution [#7723](https://entityframework.codeplex.com/SourceControl/network/forks/BrandonDahler/EntityFramework/contribution/7723)

Brandon Dahler

### DIFF
--- a/src/EntityFramework/Core/Objects/ELinq/ExpressionConverter.cs
+++ b/src/EntityFramework/Core/Objects/ELinq/ExpressionConverter.cs
@@ -109,6 +109,7 @@ namespace System.Data.Entity.Core.Objects.ELinq
 
         #region Additional Entity function names
 
+        private const string Like = "Like";
         private const string AsUnicode = "AsUnicode";
         private const string AsNonUnicode = "AsNonUnicode";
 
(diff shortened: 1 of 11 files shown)


NicoJuicy:
Please consider this in 6.2, it gives a workarround for certain edge cases where arrays of values are used in combination with .StartsWith

EF Translates this by design to CHARINDEX. When testing the same query with LIKE 'somevalue%', i had a x12 speed improvement ( from10 seconds to 0,8 seconds)

References:
https://entityframework.codeplex.com/workitem/2388
https://github.com/aspnet/EntityFramework6/issues/115

src/EntityFramework/Core/Metadata/Edm/Provider/ClrProviderManifest.cs
```diff
@@ -79,6 +79,9 @@ internal static bool TryGetPrimitiveTypeKind(Type clrType, out PrimitiveTypeKind
                     case TypeCode.Byte:
                         primitiveTypeKind = PrimitiveTypeKind.Byte;
                         break;
+                    case TypeCode.Char:
```
maumar: adding Char to the provider manifest might have significant impact in many places. Please remove this and just keep the overload that takes string escapeCharacter.

src/EntityFramework/Core/Objects/ELinq/ExpressionConverter.cs
```diff
@@ -1458,6 +1459,43 @@ private DbExpression ImplementEqualityUnknownArguments(DbExpression left, DbExpr
...
+            {
+                throw new ProviderIncompatibleException(Strings.ProviderDoesNotSupportEscapingLikeArgument);
+            }
+
+            var translatedPatternExpression = TranslateExpression(patternExpression);
+            var translatedEscapeExpression = (escapeExpression != null ? TranslateExpression(escapeExpression) : null);
+            var translatedInputExpression = TranslateExpression(inputExpression);
+
+            DbExpression result;
+            if (escapeExpression != null)
```
maumar: convert this to:
return escapeExpression != null
? translatedInputExpression.Like(translatedPatternExpression, translatedEscapeExpression)
: translatedInputExpression.Like(translatedPatternExpression);

brandondahler:
Comments addressed.

maumar:
@brandondahler new changes look good. Please rebase squash on the latest master and I will merge the pull request. Thanks!

brandondahler:
Rebased/squashed.  Did not run tests locally.

maumar:
Merged in d9f7c0fc0a0852e7c36e74c2b7b61c15e5c21712. @brandondahler thank you for the contribution and quick turnaround! Also, if you are interested in additional contribution we plan to add similar functionality to EF Core: https://github.com/aspnet/EntityFramework/issues/6159

PR #40, opened 2016-08-09
Reduce start up time by loading finished Code First models from a persistent cache

j-Edge:
Ported from [CodePlex Pull Request 8468](http://entityframework.codeplex.com/SourceControl/network/forks/jEdge/CodeFirstCache/contribution/8468)
Fix for CodePlex Work Items [#1876](http://entityframework.codeplex.com/workitem/1876) and [#2631](http://entityframework.codeplex.com/workitem/2631)
Reduce start up time by loading finished Code First models from a persistent cache
Incorporates DbModelStore persistent cache to reduce Code First Startup Time

With these changes, first AppDomain calls to context.Database.Initialize for a model with just over 600 models and a null initializer dropped from 12-14 seconds to about 1.9 seconds after the edmx was written, saving 10-12 seconds on initialization. The first call to write the edmx still ran in 12-14 seconds (no noticeable delay added).

Includes code from the patch by emilcicos attached to work item [#1876](http://entityframework.codeplex.com/workitem/1876), with modifications to fix known issues.

Code changed after applying the patch:
- DefaultDbModelStore invalidates (deletes) the persistent store based on comparing context dll and edmx file last updated timestamps.
- EdmxWriter was updated to read from the DbModelStore when available. This fixed a null reference when using DropCreateAlways initializer with a DbModelStore (compiledModel.CachedModelBuilder is null when the model is loaded from the store). Added function DbModelStore.TryGetEdmx to allow the EdmxWriter to read the edmx directly from storage.
- LazyInternalContext.ModelMatches changes in the patch were removed as they were unnecessary after the EdmxWriter fix.
- Added unit and functional test coverage for code changes including the patch code. DefaultDbModelStoreTests functional test covers the end-to-end functionality of using a DbModelStore.

### DIFF
--- a/src/EntityFramework/DbConfiguration.cs
+++ b/src/EntityFramework/DbConfiguration.cs
@@ -770,6 +770,18 @@ namespace System.Data.Entity
         }
 
         /// <summary>
+        /// Sets a singleton model store implementation (persisted model cache).
+        /// </summary>
+        /// <param name="modelStore">The model store implementation.</param>
+        protected internal void SetModelStore(DbModelStore modelStore)
+        {
+            Check.NotNull(modelStore, "modelStore");
+
+            _internalConfiguration.CheckNotLocked("SetModelStore");
+            _internalConfiguration.RegisterSingleton(modelStore);
+        }
+
+        /// <summary>
         /// Call this method from the constructor of a class derived from <see cref="DbConfiguration" /> to register
         /// a database table existence checker for a given provider.
         /// </summary>
--- a/src/EntityFramework/EntityFramework.csproj
+++ b/src/EntityFramework/EntityFramework.csproj
@@ -250,8 +250,11 @@
     <Compile Include="Infrastructure\Annotations\IndexAnnotation.cs" />
     <Compile Include="Infrastructure\Annotations\IndexAnnotationSerializer.cs" />
     <Compile Include="Infrastructure\Annotations\IndexAttributeExtensions.cs" />
+    <Compile Include="Infrastructure\DbModelStore.cs" />
+    <Compile Include="Infrastructure\DefaultDbModelStore.cs" />
     <Compile Include="Infrastructure\DependencyResolution\TransactionContextInitializerResolver.cs" />
     <Compile Include="Infrastructure\DependencyResolution\TransactionHandlerResolver.cs" />
+    <Compile Include="Infrastructure\EdmxReader.cs" />
     <Compile Include="Infrastructure\Design\AppConfigReader.cs" />
     <Compile Include="Infrastructure\Interception\DatabaseLogger.cs" />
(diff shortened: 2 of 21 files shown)


ilmax:
:rada:

brandondahler:
Instead of saving to an xml file and relying on the last modified date and other metadata, I propose we should use a resource file that automatically invalidates itself depending on whether the dll has been recompiled or not.


Sample use case where last modified date would not be sufficient:
1. Update entity assembly with new migration and deploy to production.
2. Revert change made, manually run Down on migration against database and revert .exe/.dll to previously deployed version.

Depending on deployment methodologies, it is possible that the original modification time would be restored on the .exe/.dll .  If the .entitycache was not manually deleted by revert process then our DbModelStore would load the incorrect edmx.


I wrote a basic proof of concept to allow us to read/write resources out to a XYZ.{exe,dll}.entitycache file and automatically invalidate when the executable is modified. No special assembly references are required.

``` C#
using System;
using System.Collections;
using System.IO;
using System.Linq;
using System.Reflection;
using System.Resources;

namespace Test
{
    public static class Program
    {
        public static void Main(string[] args)
        {
            var assembly = Assembly.GetExecutingAssembly();
            var currentModuleVersionId = assembly.ManifestModule.ModuleVersionId;
            Console.WriteLine(currentModuleVersionId);
            Console.WriteLine();


            var requiresRebuild = true;
            var resourcePath = assembly.Location + ".entitycache";

            if (File.Exists(resourcePath))
            {
                using (var resourceReader = new ResourceReader(resourcePath))
                {
                    var resourceDictionary = resourceReader.OfType<DictionaryEntry>()
                        .ToDictionary(de => de.Key, de => de.Value);


                    if (resourceDictionary.ContainsKey("_.Assembly.ModuleVersionId"))
                    {
                        var storedModuleVersionId = (Guid) resourceDictionary["_.Assembly.ModuleVersionId"];
                        requiresRebuild = (storedModuleVersionId != currentModuleVersionId);
                    }
                }
            }

            if (requiresRebuild)
            {
                Console.WriteLine("Rebuilding resource file.");
                using (var resourceWriter = new ResourceWriter(resourcePath))
                {
                    resourceWriter.AddResource(
                        "_.Assembly.ModuleVersionId",
                        currentModuleVersionId);
                }
            } else {
                Console.WriteLine("Rebuild not required.");
            }
        }
    }
}
```

If desired, I can provide a pull request to @j-Edge to implement this kind of cache invalidation.

ThomasZitzler:
Please keep the option to save the model to an xml file. We have a more or less dynamically building system that loads entities from multiple dlls and builds the model based on the found entities. 

There it's not possible to save it into a dll/exe as resource.

ajcvickers:
This has now been merged. Thanks for the contribution!

PR #46, opened 2016-08-13
Index/PK Fluent indexes

brandondahler:
Resolves [#1969](https://entityframework.codeplex.com/workitem/1969) and [#2163](https://entityframework.codeplex.com/workitem/2163), which requests a first-class fluent API for database indexes and the ability to configure the primary key index.

This changeset resolves these two requests by:
1. Adding a HasIndex function on the ModelConfiguration.EntityTypeConfiguration API to configure the internal EntityTypeConfiguration and return the IndexConfiguration API instance. 
2. Modifying the HasKey function to configure the internal EntityTypeConfiguration and return the PrimaryKeyIndexConfiguration API instance. 
3. Adding an abstract index configurations API and two impmenetations of that API to modify an internal index configuration. The placement of these might be incorrect/inappropriate, being in ModelConfiguration/Configuration/Properties/Index/{,Api} . 
4. Adding necessary functions to the internal EntityTypeConfiguration to store configurations for indexes and the primary key and a ConfigureIndexes function to apply those configurations to the SSpace EntityType model for the primary key and SSpace EdmProperty for the indexes. The configurations are applied by adding IndexAnnotations on the respective types. 
5. Adding an Identity property to the IndexAttribute to allow associating indexes into CompositeIndexes without specifying the index's name, as well as updating IndexAnnotationSerializer and IndexAttributeExtensions' IsCompatibleWith function to reflect this. 
6. Updating the EdmModelDiffer to handle the primary key configuration and apply it to the {,Add,Drop}PrimaryKeyOperations. 
7. Moving the IsClustered property from AddPrimaryKeyOperation to the base type and updates the inverse functions of the AddPrimaryKeyOperation and DropPrimaryKeyOperation to handle configured primary keys.

The edmx format has two changes: 
1. IndexAnnotations now have an Identity property. 
2. SSpace EntityTypes can now have an IndexAnnotation.

One final thing is that I am unsure if I should have updated any XSDs in the project. Many tests were added and modified to handle the new API. 

I believe that I have covered most of the scenarios but I'm sure I've missed some more complex or non-straightforward use-cases. 

Migrated from [CodePlex](https://entityframework.codeplex.com/SourceControl/network/forks/BrandonDahler/EntityFramework/contribution/7954).  See there for previous discussions. @ajcvickers was assigned to review.

### DIFF
--- a/src/EntityFramework.SqlServer/SqlServerMigrationSqlGenerator.cs
+++ b/src/EntityFramework.SqlServer/SqlServerMigrationSqlGenerator.cs
@@ -1613,7 +1613,10 @@ namespace System.Data.Entity.SqlServer
 
             var target = new CreateTableOperation(source.Name + suffix)
                              {
-                                 PrimaryKey = new AddPrimaryKeyOperation()
+                                 PrimaryKey = new AddPrimaryKeyOperation() 
+                                 {
+                                     IsClustered = source.PrimaryKey.IsClustered
+                                 }
                              };
 
             Debug.Assert(target.PrimaryKey.Name == source.PrimaryKey.Name + suffix);
--- a/src/EntityFramework/DataAnnotations/Schema/IndexAttribute.cs
+++ b/src/EntityFramework/DataAnnotations/Schema/IndexAttribute.cs
@@ -23,28 +23,23 @@ namespace System.ComponentModel.DataAnnotations.Schema
     [SuppressMessage("Microsoft.Performance", "CA1813:AvoidUnsealedAttributes")]
     public class IndexAttribute : Attribute
     {
-        private Guid? _identity;
         private string _name;
         private int _order = -1;
         private bool? _isClustered;
         private bool? _isUnique;
 
         /// <summary>
-        /// Creates a <see cref="IndexAttribute" /> instance for an index that will be named by convention, 
-        /// but has no identity, column order, clustering, or uniqueness specified.
+        /// Creates a <see cref="IndexAttribute" /> instance for an index that will be named by convention and
+        /// has no column order, clustering, or uniqueness specified.
         /// </summary>
         public IndexAttribute()
         {
         }
 
         /// <summary>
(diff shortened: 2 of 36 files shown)


brandondahler:
In RE: to last initial thoughts by @ajcvickers  
* I think as a team we will discuss the ConfigureKey API and other alternatives--having both HasKey and ConfigureKey could be confusing. 
  * Understood
* I am not sure about the use of guids to identify parts of indexes. I'm not sure that we need this. Could it be moved to a separate PR? 
  * Yes I can but this request would have to be made dependent on that pull request because...
  * ModelConfiguration.Configuration.Properties.Index.IndexConfiguration (and IndexAttributes in general) have to be able to relate multiple columns together.  Previously this was done only based on IndexAttribute.Name; however, I can't assign an arbitrary name within IndexConfiguration.  After going through the different possibilites, I decided that generating GUIDs would be the only truly safe way.
* It looks like you removed IsClustered from AddprimaryKeyOperation. This would be a breaking change, and I'm pretty sure we can't accept any breaking changes for 6.2, so I don't think this can be removed
  * It was not truly removed from AddPrimaryKeyOperation but instead it got moved to its base class (PrimaryKeyOperation).

ajcvickers:
@brandondahler Sorry for taking so long to come back to this. With regard to the GUIDs and naming, you say that you can't assign an arbitrary name, but if no name is given for an index, then we already generate a name. Currently, this is done in Migrations, but I think it could be done as part of the fluent API implementation. We would need to do uniquification for name collisions, but we do that for other things we name so I don't think it will be a major problem. That would then allow the GUID identities to be removed.

ajcvickers:
@brandondahler Just a friendly ping on this before it becomes too late for 6.2.

brandondahler:
@ajcvickers Updates made, let me know what you think and I'll rebase/squash if desired.

ajcvickers:
This has now been merged. Thanks for the contribution!

PR #47, opened 2016-08-13
Unblock model defined functions

brandondahler:
Work Item [#2615](https://entityframework.codeplex.com/workitem/2615) - Unblock creation of model defined functions in conventions 
- Updated CSDL schema to handle existing serialization of model defined functions.
- Removed outdated Debug.Assert and DebugCheck.NotNull. 
- Added tests for new functionality.

Migrated from [CodePlex](https://entityframework.codeplex.com/SourceControl/network/forks/BrandonDahler/EntityFramework/contribution/8302).  Originally assigned to @AndriySvyryd .


brandondahler:
In RE: to:
* The changes look good, but it would be better to have more tests to ensure all the attributes are exercised,
  * I would be happy to do the implementation, but I do not have enough understanding of what the different attributes represent to create useful tests.

PR #67, opened 2016-09-20
Fix AddOrUpdate for nullable identifiers

### LINKED ISSUE #9: AddOrUpdate error when identifyingProperty is nullable
Hi,

There is an error in AddOrUpdate method. Let's look on such example:

```
public class Test
{
    public Guid? SomeKey { get; set;}
}

protected override void Seed(TestContext context)
{
        var entity = new Test() { SomeKey = Guid.NewGuid()};
        context.Tests.AddOrUpdate(t => t.SomeKey, entity); 
}
```

Running such code creates error:

> The binary operator Equal is not defined for the types 'System.Nullable`1[System.Guid]' and 'System.Guid'.

Problem lays in invalid condition createion based on identifier parameter in AddOrUpdate method:

```
Expression.Equal(
                            Expression.Property(parameter, pi.Single()),
                            Expression.Constant(pi.Last().GetValue(entity, null)))
```

Expression.Property returns expression of type Guid?. Expression.Constant return expression of type Guid. And Equals cannot join it.

The solution is to add parameter to Expression.Constant about type of property.

```
                            Expression.Constant(pi.Last().GetValue(entity, null), pi.Last().PropertyType))
```

Then it will cast this expression to be Guid? and it will run property.

I think that this will also occure on EF7. but i didn't use it already.

axelheer:
_Note:_ many tests fail on my machine. Thus, I only verified those within `DbSetMigrationsExtensions`. Maybe my modification of `MigrationsCustomer` has some unwanted side effect regarding other tests?

Fixes #9

### DIFF
--- a/src/EntityFramework/Migrations/DbSetMigrationsExtensions.cs
+++ b/src/EntityFramework/Migrations/DbSetMigrationsExtensions.cs
@@ -133,7 +133,7 @@ namespace System.Data.Entity.Migrations
                     = identifyingProperties.Select(
                         pi => Expression.Equal(
                             Expression.Property(parameter, pi.Single()),
-                            Expression.Constant(pi.Last().GetValue(entity, null))))
+                            Expression.Constant(pi.Last().GetValue(entity, null), pi.Last().PropertyType)))
                                            .Aggregate<BinaryExpression, Expression>(
                                                null,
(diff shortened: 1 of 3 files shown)


axelheer:
cc @anpete

PR #87, opened 2016-10-14
Implement script generation support in migrate.exe - fixes codeplex/251

Sebazzz:
As per Codeplex [issue#251](http://entityframework.codeplex.com/workitem/251), implemented support for script generation via Entity Framework' migrate.exe.

This enables build automation regarding migration script, for example, generate migration scripts in the build server to include in a deployment. 

As part of this commit, in order to add new resources, `Resources.Migrate.cs` is also re-generated. Since the generated code is apparently out-of-date since the last time the T4 file was changes, the changes in that file are somewhat bigger than I'd like them to be.

### DIFF
--- a/src/Migrate/Program.cs
+++ b/src/Migrate/Program.cs
@@ -8,6 +8,7 @@ namespace System.Data.Entity.Migrations.Console
     using System.Diagnostics;
     using System.IO;
     using System.Reflection;
+    using System.Text;
     using CmdLine;
     using Console = System.Console;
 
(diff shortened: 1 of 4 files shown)


src/Migrate/Program.cs
```diff
@@ -77,10 +77,24 @@ public void Run()
...
+                    facade.Update(_arguments.TargetMigration, _arguments.Force);
+                }
             }
         }
 
+        private static void ScriptUpdate(ToolingFacade facade, Arguments arguments)
+        {
+            string scriptContents = facade.ScriptUpdate(arguments.SourceMigration, arguments.TargetMigration, arguments.Force);
+
+            File.WriteAllText(arguments.ScriptFile, scriptContents);
```
ErikEJ: Consider adding Encoding.UTF8

IDisposable:
:shipit:

anpete:
Thanks @Sebazzz.

PR #126, opened 2016-12-06
Update tests to current default LocalDb instance

rowanmiller:
The LocalDb tests currently depend on (localdb)\v11.0 which was the default instance name created when installing an earlier version of VS. Nowadays the instance is (localdb)\mssqllocaldb and should remain that way for future versions of VS.
All tests in SimpleScenariosForLocalDb now pass on a machine without v11.0.

I've already updated our CI to have the new instance name available.


ajcvickers:
v11.0 also appears in DatabaseTests

rowanmiller:
Yeah I just found that while running the tests... not sure how I missed it with find/replace 😄

PR #136, opened 2016-12-09
PowerTools: Upgrade to a v2 vsixmanifest

bricelam:
This drops support for VS 2010, but adds support for VS 2017 and future releases (in theory).

cc @divega @rowanmiller


bricelam:
Hmm...
image

bricelam:
It doesn't appear to be working in VS 2017 RC. This PR may not be worth taking.

lajones:
Agreed. See the email I just sent round about why this won't work for VS15 (basically we will need VSIX V3 for that) - so I think we should take the other approach for now.

bricelam:
Recalling.

Issue #139, opened 2016-12-09
PowerTools: Support for non-default output path when using |datadirectory| in connection strings


divega:
(Originally reported at https://entityframework.codeplex.com/workitem/2428. I am reviving here because I have the fix and it is trivial) 

Setting an output path that not a child of the directory containing the project will result in a target directory that contains path traversals, e.g. `C:\EfTestProg\EfTestProg\..\Output\Debug\` but SqlClient isn't happy with those. Fix is to add a call to `Path.GetFullPath()` when computing the data directory.

PR #140, opened 2016-12-09
Fix for #139: PowerTools: setting |datadirectory| path fails if using non-default output path

divega:
Setting an output path that not a child of the directory containing the project will result in a target directory that contains path traversals, e.g. C:\EfTestProg\EfTestProg\..\Output\Debug\ but SqlClient isn't happy with those. Fix is to add a call to Path.GetFullPath() when computing the data directory.

Verification: smoke testing.

### DIFF
--- a/src/PowerTools/DbContextPackage.cs
+++ b/src/PowerTools/DbContextPackage.cs
@@ -518,9 +518,10 @@ namespace Microsoft.DbContextPackage
 
             AppDomain.CurrentDomain.SetData(
                 "DataDirectory",
-                project.IsWebProject()
-                    ? Path.Combine(project.GetProjectDir(), "App_Data")
-                    : project.GetTargetDir());
+                Path.GetFullPath(
+                    project.IsWebProject()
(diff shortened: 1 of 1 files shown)


src/PowerTools/DbContextPackage.cs
```diff
@@ -520,7 +520,7 @@ private static void SetDataDirectory(Project project)
                 "DataDirectory",
                 project.IsWebProject()
                     ? Path.Combine(project.GetProjectDir(), "App_Data")
-                    : project.GetTargetDir());
+                    : Path.GetFullPath(project.GetTargetDir()));
```
bricelam: Should it be around the whole ternary expression? (Or even in `GetTargetDir` before it returns)

divega: I am not sure it will ever be necessary for WebSites but I can doing over the whole ternary expression. 

I could do it on the extension methods but setting the data directory is the only place where I know path traversals are not supported.

PR #141, opened 2016-12-09
Update projects to .NET 4.5 and rename to Beta 5.

lajones:
@bricelam @divega This is just the update to make the .csproj's target NET4.5 and to update the name to Beta 5. This is just to get the dev environment baselined so it will work again. After that I can get going on fixes.


bricelam:
Should [this](https://github.com/aspnet/EntityFramework6/blob/d1a630f4a309fec944f4dd361314c9b22e4bf89e/src/PowerTools/source.extension.vsixmanifest#L38) also be updated?

test/PowerTools.Test/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Moq" version="4.0.10827" />
+  <package id="Moq" version="4.0.10827" requireReinstallation="True" />
```
bricelam: Revert this.

bricelam: Whenever you change target framework, NuGet flags them for reinstall, but this version of Moq only has a net40 library, so it's not required.

lajones: This was put in automatically when I updated to NET 4.5. But I'm not tied to it so if you think we need to get rid of it so be it.

lajones: This was put in automatically when I updated to NET 4.5. But I'm not tied to it so if you think we need to get rid of it so be it.

lajones:
Re your other comment - I don't think that needs to change. I mentioned that it still works on Dev10 and I think that may be part of the reason. And Julie found no need to change that line for Dev15.

lajones:
Checked into master branch with #e266556f.

Issue #22, opened 2016-07-29
DatabaseLogFormatter logs incorrect sql query execution times


rowanmiller:
If you new up multiple DbContext and make queries in parallel, each context's DatabaseLogFormatter receives interception notifications of every other DbContext! 
 It confuses the DatabaseLogFormatter, which seems to have been written without that scenario in mind. 
 Therefore DatabaseLogFormatter will stop its StopWatch even though the event comes from the query from another context. 
 I guess it's a bug at the interception level. A DbContext should only received interceptions events relating to queries/commands issued against that particular context. 
 Sorry I don't have time to build a repro, but if that's not clear I can provide additional details...

rowanmiller:
Ported from CodePlex https://entityframework.codeplex.com/workitem/2759

PR #153, opened 2016-12-14
Use separate timer per interception context when DatabaseLogFormatter is used without a context

ajcvickers:
Issue #22

In the most common use, the DatabaseLogFormatter is used for a single DbContext, which means that there can be no interleaving between starting and finishing an execution. However, if the formatter is being used for multiple contexts, then these calls can be interleaved, which means that each pair needs its own timing.

Since each call has its own interception context, this can be used to store the timer. We accepted a pull request to store this kind of data, but I realized that if there are multiple interceptors and each want to use this data slot, the result would be one clobbering the other. Therefore, I obsoleted that mechanism in favor of a mechanism that can be used by multiple interceptors.

The change required that Stopwatch be obsoleted. However, the existing functionality has been retained so that an in-place update or ignoring of the obsolete warning will not make working code break.

### DIFF
--- a/src/EntityFramework/Infrastructure/Interception/DatabaseLogFormatter.cs
+++ b/src/EntityFramework/Infrastructure/Interception/DatabaseLogFormatter.cs
@@ -30,6 +30,7 @@ namespace System.Data.Entity.Infrastructure.Interception
     /// </remarks>
     public class DatabaseLogFormatter : IDbCommandInterceptor, IDbConnectionInterceptor, IDbTransactionInterceptor
     {
+        private const string StopwatchStateKey = "__LoggingStopwatch__";
         private readonly WeakReference _context;
         private readonly Action<string> _writeAction;
         private readonly Stopwatch _stopwatch = new Stopwatch();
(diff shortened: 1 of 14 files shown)


src/EntityFramework/Infrastructure/Interception/DatabaseLogFormatter.cs
```diff
@@ -30,6 +30,7 @@ namespace System.Data.Entity.Infrastructure.Interception
     /// </remarks>
     public class DatabaseLogFormatter : IDbCommandInterceptor, IDbConnectionInterceptor, IDbTransactionInterceptor
     {
+        private const string StopwatchStateKey = "__LoggingStopwatch__";
         private readonly WeakReference _context;
         private readonly Action<string> _writeAction;
         private readonly Stopwatch _stopwatch = new Stopwatch();
```
AndriySvyryd: Shouldn't this be initialized lazily now in case it's not used?

ajcvickers: @AndriySvyryd It is actually still always used to preserve back compatibility with code that has overridden some but not all methods.

PR #161, opened 2016-12-22
Fix to #6 - .NET/EF6 ObjectContext leak using compiled LINQ query that references String.StartsWith()

maumar:
Problem was that when translating StartsWith/EndsWith/Contains into Like, we are escaping the argument in the runtime using ProviderManifest. Originally we were doing this by adding Func rooted in ExpressionConverter as part of the QueryParameterExpression. ExpressionConverter in turn holds reference to the context which caused the original problem (since the parameter is being added to the cache, and hence persists through the garbage collection)

Fix is to sever the connection to ExpressionConverter by using Expression<Func<>> instead and pass ProviderManifest a constant, rather than closure. Also, original method contained out parameter, which now have been converted to a Tuple<> result.

### DIFF
--- a/src/EntityFramework/Core/Objects/ELinq/ExpressionConverter.cs
+++ b/src/EntityFramework/Core/Objects/ELinq/ExpressionConverter.cs
@@ -1518,10 +1518,19 @@ namespace System.Data.Entity.Core.Objects.ELinq
             if (providerSupportsEscapingLikeArgument && (queryParameterExpression != null))
             {
                 useLikeTranslation = true;
-                bool specifyEscapeDummy;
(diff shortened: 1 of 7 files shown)


maumar:
@anpete @divega

anpete:
Test?

maumar:
This only affects ObjectContext compiled query - we don't have test infrastructure for compiled query tests in EF6 codebase

anpete:
@divega Thoughts on spending time adding a test for this? I think it would be nice, but not essential (I'm assuming @maumar has done manual verification).

maumar:
new version up - added test

PR #172, opened 2017-01-17
Fix #13 - Migrate.exe runs all the migrations even if they are alread…

### LINKED ISSUE #13: Migrate.exe runs all the migrations even if they are already applied
It looks like migrate.exe is not interpreting right the exceptions - for example if user, that is used to connect to database when running migrate.exe does not have rights to read migration history table (is only ddladmin), migrate.exe tries to run all the migrations, instead of showing error for missing permissions. 

 It took lots of time and reverse engineering to figure this out.

anpete:
…y applied

- Adds simple detection of permission denied DB error. Provided hook is added to MigrationSqlGenerator for simplicity instead of introducing a new seam.


anpete:
@bricelam

Issue #176, opened 2017-01-19
Length of string parameters on where clause


frbatist:
I have the following mapping for a varchar column:

`modelBuilder.Entity<Empresa>().Property(p => p.Codigo).HasColumnName("EMPCOD").HasMaxLength(4).IsFixedLength();`

And I'm ending up with a parameter like this one:
@p__linq__0 varchar(8000)

Is there any chance to fix this so that the parameter is created the wright way ?
Like this :
@p__linq__0 varchar(4)

rowanmiller:
**EF Team Triage:** This issue is not something that our team is planning to address in the EF6.x code base.  This does not mean that we would not consider a community contribution to address this issue.

Moving forwards, our team will be fixing bugs, implementing small improvements, and accepting community contributions to the EF6.x code base. Larger feature work and innovation will happen in the EF Core code base (https://github.com/aspnet/EntityFramework).

**Closing an issue in the EF6.x project does not exclude us addressing it in EF Core. In fact, a number of popular feature requests for EF have already been implemented in EF Core (alternate keys, batching in SaveChanges, etc.).**

*BTW this is a canned response and may have info or details that do not directly apply to this particular issue. While we'd like to spend the time to uniquely address every incoming issue, we get a lot traffic on the EF projects and that is not practical. To ensure we maximize the time we have to work on fixing bugs, implementing new features, etc. we use canned responses for common triage decisions.*

divega:
PR was submitted for this: https://github.com/aspnet/EntityFramework6/pull/181/. Reopening so that we can evaluate it.

rowanmiller:
Putting in 6.3.0 on the basis of the PR. It's not something our team is planning to do.

ajcvickers:
Closing for now. Will re-open if the PR gets to a state in which it can be merged.

PR #181, opened 2017-01-23
Issue #176 Length of string parameters on where clause

frbatist:
I have made some changes to solve the issue 176, just made the same thing that is done for non-unicode strings (varchar vs nvarchar). I get the MaxLength facet from the dbType, and create the TypeUsage with it.


ajcvickers:
@frbatist Doesn't this change the behavior if the actual string data doesn't fit in the specified max length? EF Core handles a similar thing in its type mappings, but it does it on a per parameter instance that takes into account that actual length of data being passed. This is important because the specified max length cannot be relied upon as being the max length actually configured in the database.

Issue #165, opened 2017-01-06
EF6 does not work with primary key from sequence


Dixin:
If a table with primary key from sequence instead of identity, for example:
```sql
CREATE TABLE [Application].[Countries](
    [CountryID] [int] NOT NULL
    CONSTRAINT [PK_Application_Countries] PRIMARY KEY CLUSTERED
)
GO

ALTER TABLE [Application].[Countries] ADD  CONSTRAINT [DF_Application_Countries_CountryID]  DEFAULT (NEXT VALUE FOR [Sequences].[CountryID]) FOR [CountryID]
GO
```
The the entity would be:
```c#
public class Country
{
    [Key]
    [DatabaseGenerated(???)]
    public int CountryID { get; set; }
}
```
With code first, EF6 can query, but cannot insert/update entity.
* If `DatabaseGeneratedOption.Identity` is specified, EF6 generates `scope_identity()` which fails
* If `DatabaseGeneratedOption.Computed` is specified, EF6 throws `DbUpdateException`: Modifications to tables where a primary key column has property 'StoreGeneratedPattern' set to 'Computed' are not supported. Use 'Identity' pattern instead. Key column: 'CountryID'. Table: 'CodeFirstDatabaseSchema.Countries'.
* If `DatabaseGeneratedOption.None` is specified, EF6 does not refresh the entity's primary key after inserting it to table.

I searched around, and couldn't find a solution. Any workaround for this? And any plan to support sequence in EF6?

rowanmiller:
@divega may have a changeset that already addresses this, which we can look at taking.

In the meantime, one potential workaround is to map to stored procedures so that you can control the INSERT SQL.

divega:
Clearing up milestone to consider enabling to workaround this in 6.2 as per discussion in https://github.com/aspnet/EntityFramework6/pull/182.

divega:
A workaround for this issue was added in 8e38ed8e4add6d4df1d6f1fa9013fa3d2c6dd84d. It requires setting the following flag in the application:

``` C#
SqlProviderServices.UseScopeIdentity = false;
```

Here is the description of the new flag (which includes some caveats worth knowing):

[...]

PR #182, opened 2017-01-23
Fix for #165: Add flag to control whether scope_identity() is used

divega:
… to retrieve values generated for numeric columns during insert operations.

Currently the logic is hardcoded in the `DmlSqlGenerator` class of the SQL
Server provider for EF6, in particular in the
`IsValidScopeIdentityColumnType()` method to use `scope_identity()` to
recover generated values for any columns for which it may apply:

``` SQL
INSERT [dbo].[People]([Name])
VALUES (@0)
SELECT [Id]
FROM [dbo].[People]
WHERE @@ROWCOUNT > 0 AND [Id] = scope_identity()
```

This PR adds a flag UseScopeIdentity with a default value of true in
SqlProviderServices (in the past we added `TruncateDecimalsToScale` in a
similar fashion). If the flag is set to false, the following SQL will be
generated which uses the OUTPUT clause:

``` SQL
DECLARE @generated_keys table([Id] int)
INSERT [dbo].[People]([Name])
OUTPUT inserted.[Id] INTO @generated_keys
VALUES (@0)
SELECT t.[Id]
FROM @generated_keys AS g JOIN [dbo].[People] AS t ON g.[Id] = t.[Id]
WHERE @@ROWCOUNT > 0
```

Verification:

@ajcvickers ran all tests on a bootleg that defaults the value of `UseScopeIdentity` to false. The only failures were on tests that directly assert on that the generated SQL for inserts matches the default behavior:

`System.Data.Entity.SqlServer.SqlGen.DmlFunctionSqlGeneratorTests.Insert_simple_tph_entity_base`
`System.Data.Entity.SqlServer.SqlGen.DmlFunctionSqlGeneratorTests.Insert_simple_entity`
`System.Data.Entity.SqlServer.SqlGen.DmlFunctionSqlGeneratorTests.Insert_simple_tph_entity_derived`
`System.Data.Entity.Interception.CommandInterceptionTests.Multiple_contexts_running_concurrently_can_use_interception`


divega:
This is the general approach I had in mind. Thoughts?

AndriySvyryd:
In general this sounds fine, but making this a per-entity type configuration would be better to limit the perf impact.

divega:
@AndriySvyryd it is just about 5.11 times slower for a simple Person(int, string) case :smile_cat:

ajcvickers:
@divega I think it would be great if this could be done with an annotation on specific columns, but I think that if this is what we have time to do, then it is worth doing this because it will unblock the main scenarios.

divega:
@ajcvickers Agreed. We could consider is having something like this in 6.2. It shouldn't prevent us from having something more granular for 6.3.

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -92,6 +92,7 @@ private SqlProviderServices()
                                && asSpatialKey.ProviderInvariantName == ProviderInvariantName
                                && SupportsSpatial(asSpatialKey.ProviderManifestToken);
                     }));
+            UseScopeIdentity = true;
```
divega: Will add an empty line before I push.

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -132,6 +133,15 @@ public static bool TruncateDecimalsToScale
         }
 
         /// <summary>
+        /// Set this flag to false to use an OUTPUT clause instead of the SCOPE_IDENTITY() function to 
+        /// retrieve values generated on the database for numeric columns during an INSERT operation. 
+        /// This makes it possible to retrieve values generated by sequences or other means, but setting 
+        /// the flag to the default value of true is recommended and can provide better performance if all 
+        /// values are generated using IDENTITY columns.
+        /// </summary>
```
divega: @rowanmiller can you review the doc comment?

rowanmiller: This would align better with how we've written docs for other flags...

> Gets or sets a value indicating whether to use the SCOPE_IDENTITY() function retrieve values generated by the database for numeric columns during an INSERT operation. The default value of <c>true</c> is recommended and can provide better performance if all numeric values are generated using IDENTITY columns. If set to <c>false</c>, an OUTPUT clause will be used instead. An OUTPUT clauses makes it possible to retrieve values generated by sequences, or other means.

rowanmiller: BTW, should wrap true and false in c tage...  `<c>true</c>` and `<c>false</c>`

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -115,23 +117,35 @@ public static SqlProviderServices Instance
...
-        /// the scale (number of decimal places) defined for the column. The default value is true,
-        /// indicating that decimal values will be truncated, in order to prevent breaking existing
-        /// applications that depend on this behavior.
-        /// </summary>
-        /// <remarks>
-        /// With this flag set to true <see cref="SqlParameter" /> objects are created with their Scale
-        /// properties set. When this flag is set to false then the Scale properties are not set, meaning
-        /// that the truncation behavior of SqlParameter is avoided.
-        /// </remarks>
+        /// Get or sets a value indicating whether <see cref="decimal" /> parameter values are truncated to
```
divega: Change this too to align with the structure recommended by @rowanmiller. 

cc @ajcvickers

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -115,23 +117,35 @@ public static SqlProviderServices Instance
...
+        /// The default value is <c>true</c> to prevent breaking existing applications that depend on this 
+        /// behavior. 
         public static bool TruncateDecimalsToScale
         {
             get { return _truncateDecimalsToScale; }
             set { _truncateDecimalsToScale = value; }
         }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to use the SCOPE_IDENTITY() function to retrieve values
```
divega: Updated based on @rowanmiller's feedback.

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -115,23 +117,35 @@ public static SqlProviderServices Instance
...
         {
             get { return _truncateDecimalsToScale; }
             set { _truncateDecimalsToScale = value; }
         }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to use the SCOPE_IDENTITY() function to retrieve values 
+        /// generated by the database for numeric columns during an INSERT operation. The default value of 
+        /// <c>true</c> is recommended and can provide better performance if all numeric values are generated using 
+        /// IDENTITY columns. If set to <c>false</c>, an OUTPUT clause will be used instead. An OUTPUT clauses
```
rowanmiller: Just noticed a typo (presumably my typo 😄), "An OUTPUT clauses..." should be "An OUTPUT clause..."

PR #184, opened 2017-01-25
Fix #73 - Evaluation of DbQuery<T>.ToString() frequently times out in…

### LINKED ISSUE #73: Evaluation of DbQuery<T>.ToString() frequently times out in the debugger
`DbQuery<T>.ToString()` was overridden in order to provide a simple means to obtain the SQL translation for a LINQ query, both programmatically and at debug time. When it works property, it provides a very nice experience. 

However `DbQuery<T>.ToString()` has been found to be one of the top sources of timeouts while the debugger performs function evaluation. In general, functions that take a long time to evaluate can lead to a very lousy user experience using VS:
- During function evaluation, the debugger UI is blocked
- Once the evaluation of a function times out, the debugger preventively blocks any further function evaluation

We believe the main reasons `DbQuery<T>.ToString()` can take a long time to evaluate (and possibly exceed function evaluation timeout period) are:
- The first query executed over a particular model may require model and database initialization, which usually involves expensive computation and several database roundtrips
- A complex query may take a long time to compile

The VS debugger team has recommended that we make changes in future versions of EF6.x. After discussing several options with them, there doesn't seem to be any way to preserve the optimal simplicity of the current experience for the best case scenario (when the method returns in a timely manner) and at the same time we improve the experience for the worst case scenario (when the method takes too long to evaluate). 

In summary, the proposed changes are:

a. Add a read-only `Sql` property on `DbQuery<T>` which invokes `ToString()` and exposes the result 
b. Add a call to `Debugger.NotifyOfCrossThreadDependency()` on model and database initialization 
c. Possibly add `Debugger.NotifyOfCrossThreadDependency()` calls in expensive (but not reentrant or repetitively invoked) parts of query compilation, or alternatively use some heuristic to estimate the query complexity and invoke the method if the query falls above certain threshold.
d. Annotate `DbQuery<T>` (and its derived types) with an attribute that restores the default behavior for the debugger, e.g. something like `[DebuggerDisplay(@"\{System.Data.Entity.Infrastructure.DbQuery\}")]` (this is oversimplified, doesn't include the generic argument). Alternatively we could include a message there pointing to the new `Sql` property.

**Brief explanation:**
Function evaluation works by hijacking a thread from the application that is paused to execute code. `Debugger.NotifyOfCrossThreadDependency()` can be used in a property to indicate the debugger that the code is going down some slow path so it can abort the evaluation in a more orderly manner than with a timeout. E.g. on the locals window, it shows a message saying "The function evaluation requires all threads to run." and presents a "threads" glyph that users can click on to force the evaluation. 

There is no way to get the glyph on the root node of an instance or on a method.

The resulting experience would look like this for a query `x`:

capture

anpete:
… the debugger

- Adds a read-only Sql property on DbQuery<T> which invokes ToString() and exposes the result
- Adds a call to Debugger.NotifyOfCrossThreadDependency() on model and database initialization
- Annotates DbQuery<T> with an attribute that restores the default behavior for the debugger, I.e. calling base.ToString()

### DIFF
--- a/src/EntityFramework/Infrastructure/DbQuery.cs
+++ b/src/EntityFramework/Infrastructure/DbQuery.cs
@@ -6,6 +6,7 @@ namespace System.Data.Entity.Infrastructure
     using System.ComponentModel;
     using System.Data.Entity.Internal.Linq;
     using System.Data.Entity.Resources;
+    using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using System.Linq.Expressions;
@@ -15,6 +16,7 @@ namespace System.Data.Entity.Infrastructure
     /// </summary>
     [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix")]
     [SuppressMessage("Microsoft.Design", "CA1010:CollectionsShouldImplementGenericInterface")]
+    [DebuggerDisplay(@"{DebuggerDisplay()}")]
     public abstract class DbQuery : IOrderedQueryable, IListSource, IInternalQueryAdapter
 #if !NET40
 , IDbAsyncEnumerable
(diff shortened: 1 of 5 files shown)


src/EntityFramework/Infrastructure/DbQuery.cs
```diff
@@ -217,7 +219,21 @@ public DbQuery<TElement> Cast<TElement>()
...
+        private string DebuggerDisplay()
+        {
+            return base.ToString();
+        }
+
+        /// <summary>
+        /// Returns the SQL that will be run against the underlying data store when this query is executed.
+        /// </summary>
+        /// <returns> The SQL string. </returns>
+        public string Sql
```
anpete: @divega Should we call this "DebugSql" to line up with Expression.DebugView?

divega: I don't feel strongly, but the name `Sql` seems correct and unambiguous enough to me.

PR #188, opened 2017-02-02
Enables building EFTools using Visual Studio 2017

divega:
Fixes all the issues I have encountered so far with running `BuildEFTools.cmd /t:build` on a Developer Command Prompt for VS 2017 RC.3. After this I still get a couple of warnings when building EntityFramework.PowerShell.csproj (but this could be unrelated):

> C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\bin\Microsoft.Common.CurrentVersion.targets(2493,5): warning MSB3283: Cannot find wrapper assembly for type library "EnvDTE
". Verify that (1) the COM component is registered correctly and (2) your target platform is the same as the bitness of the COM component. For example, if the COM component is 32-bit, your target pla
tform must not be 64-bit. [C:\Users\dbveg\Documents\GitHub\EntityFramework6\src\EntityFramework.PowerShell\EntityFramework.PowerShell.csproj]
C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\bin\Microsoft.Common.CurrentVersion.targets(2493,5): warning MSB3283: Cannot find wrapper assembly for type library "VSLang
Proj". Verify that (1) the COM component is registered correctly and (2) your target platform is the same as the bitness of the COM component. For example, if the COM component is 32-bit, your target
 platform must not be 64-bit. [C:\Users\dbveg\Documents\GitHub\EntityFramework6\src\EntityFramework.PowerShell\EntityFramework.PowerShell.csproj]

### DIFF
--- a/src/Strict.ruleset
+++ b/src/Strict.ruleset
@@ -5,5 +5,9 @@
     <Rule Id="CA1062" Action="None" />
     <Rule Id="CA2243" Action="None" />
     <Rule Id="CA1809" Action="None" />
+    <Rule Id="CA1506" Action="None" />
+    <Rule Id="CA1020" Action="None" />
+    <Rule Id="CA2210" Action="None" />
+    <Rule Id="CA1305" Action="None" />
   </Rules>
 </RuleSet>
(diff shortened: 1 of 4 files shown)


src/Strict.ruleset
```diff
@@ -5,5 +5,9 @@
     <Rule Id="CA1062" Action="None" />
     <Rule Id="CA2243" Action="None" />
     <Rule Id="CA1809" Action="None" />
+    <Rule Id="CA1506" Action="None" />
```
divega: Not sure we actually want any of these. We can discuss.

divega:
Superseded by https://github.com/aspnet/EntityFramework6/pull/200.

Issue #190, opened 2017-02-02
EF6.2 API Review Changes


rowanmiller:
- [x] Determine if moving `IsClustered` from `AddPrimaryKeyOperation` to `PrimaryKeyOperation` is a binary breaking change
- [x] Rename `DefaultDbModelStore.Location` to `Directory` (and constructor parameter too)
- [x] in `EntityTypeConfiguration<TEntityType>` replace the `ConfigureKey()` method with nested closure style overloads of `HasKey(...)`
- [x] Remove `IndexConfigurationBase<TSelf>` (probably just move the two APIs onto it's derived types)
- [x] Try to remove the public API additions from the "exceeded maximum parameters" fix (`EntityParameterCollection.GetParameterByValue(string value)` etc.)

rowanmiller:
@divega will handle the last bullet point.

PR #193, opened 2017-02-03
Fixes #185 by moving the optimization for update model from database into tools code

### LINKED ISSUE #185: Fix in PR #41 for updating large models from database has additional issues
https://github.com/aspnet/EntityFramework6/pull/41 very ingeniously fixes issues with large models by consolidating parameters of schema discovery queries that contain the same values.

I have observed a few issues with it though:

1. Modifications and additional public surface on runtime pieces like EntityClient seem unnecessary
2. We would need to verify that this works with third party providers: 
 - SQL Server is the only provider for which we know 100% the fix is needed 
 - Any provider that implements schema discovery by faking the execution of schema discovery queries may have a hard dependency with the position of the parameters in the parameter collection

divega:
Supersedes https://github.com/aspnet/EntityFramework6/pull/41 with a very similar approach (avoiding creating duplicate parameters with the same values) but the implementation in this case lives completely into the EF Tools code and avoids changes to the EF Runtime.


lajones:
Looks good to me. I like the Dictionary approach.

divega:
Also addressed last bullet point of #190.

Issue #162, opened 2016-12-26
problem with OFFSET and FETCH when the order is by a column that has one value


luizfbicalho:
If I have a class like this for example

`class Entity{
    int Id {get;set;}
    string Description {get;set;}
    int Size {get;set;} // size can be 1,2 or 3
}`

and I query this data in a grid with order and pagins ( in my case I'm using kendo ui) and I order by size, and filter where the size == 1, I have problems when I change the pages, the data that is returned is the same every page.

the sql is like this

SELECT * 
FROM Entity
WHERE Size = 1
ORDER BY Size
OFFSET 10 ROWS 
FETCH NEXT 10 ROWS ONLY

and to solve this problem it should be like 

SELECT * 
FROM Entity
WHERE Size = 1
ROW_NUMBER() over (ORDER BY Size ASC) 
OFFSET 10 ROWS 
FETCH NEXT 10 ROWS ONLY

this is caused by this feature, if it's not possible to change the order, ate least I need a way to change the sql genereator.

https://data.uservoice.com/forums/72025-entity-framework-feature-suggestions/suggestions/2541081-use-offset-and-fetch-for-skip-and-take-in-sql-serv

ErikEJ:
You can force a 2008 manifest token http://stackoverflow.com/questions/4741499/how-to-configure-providermanifesttoken-for-ef-code-first

maumar:
@luizfbicalho could you provide some more details about your scenario (model used, listing of sql that is being issues, database used etc). I tried the following:

```cs
        static void Main(string[] args)
        {
            Database.SetInitializer(new DropCreateDatabaseAlways<MyContext>());
            using (var ctx = new MyContext())
            {
                for (int i = 0; i < 50; i++)
                {
                    var e = new MyEntity { Description = "d" + i, Size = 1 };
                    ctx.Entities.Add(e);
                }

                ctx.SaveChanges();
            }

            using (var ctx = new MyContext())
            {
                ctx.Database.Log = Console.WriteLine;
                var batch1 =  ctx.Entities.OrderBy(e => e.Size).Where(e => e.Size == 1).Skip(10).Take(10).ToList();
                var batch2 = ctx.Entities.OrderBy(e => e.Size).Where(e => e.Size == 1).Skip(20).Take(10).ToList();
            }
        }
```

[...]

PR #201, opened 2017-02-13
[6.2] Fix to #162 - problem with OFFSET and FETCH when the order is by a column that has one value

maumar:
Problem affects some queries using paging that use ambiguous ordering. Fix is to introduce ORDER BY row_number() to remove the ambiguity.

New method is still able to take advantage of indexes, so there shouldn't be a significant impact on performance.

### DIFF
--- a/src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
+++ b/src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
@@ -2552,7 +2552,16 @@ namespace System.Data.Entity.SqlServer.SqlGen
                 input.Select.Skip = new SkipClause(HandleCountExpression(e.Count));
 
                 // Add the ORDER BY part.
-                AddSortKeys(input.OrderBy, e.SortOrder);
+                if (SqlProviderServices.UseRowNumberOrderingInOffsetQueries)
+                {
+                    input.OrderBy.Append("row_number() OVER (ORDER BY ");
+                    AddSortKeys(input.OrderBy, e.SortOrder);
+                    input.OrderBy.Append(")");
+                }
+                else
+                {
+                    AddSortKeys(input.OrderBy, e.SortOrder);
+                }
 
                 symbolTable.ExitScope();
                 selectStatementStack.Pop();
--- a/src/EntityFramework.SqlServer/SqlProviderServices.cs
+++ b/src/EntityFramework.SqlServer/SqlProviderServices.cs
@@ -100,6 +100,7 @@ namespace System.Data.Entity.SqlServer
         private static readonly SqlProviderServices _providerInstance = new SqlProviderServices();
 
         private static bool _truncateDecimalsToScale = true;
+        private static bool _useRowNumberOrderingInOffsetQueries = true;
 
         /// <summary>
         /// The Singleton instance of the SqlProviderServices type.
@@ -129,6 +130,16 @@ namespace System.Data.Entity.SqlServer
         {
             get { return _truncateDecimalsToScale; }
             set { _truncateDecimalsToScale = value; }
+        }
+
+        /// <summary>
+        /// Set this flag to false to prevent ordering using a row_number() function for queries with OFFSET operation, and order by columns directly instead.
+        /// </summary>
+        /// <remarks>Using row_number() can reduce ambiguity of the ordering result for some cases.</remarks>
+        public static bool UseRowNumberOrderingInOffsetQueries
+        {
+            get { return _useRowNumberOrderingInOffsetQueries; }
+            set { _useRowNumberOrderingInOffsetQueries = value; }
         }
 
         /// <summary>
--- a/test/EntityFramework/FunctionalTests/PlanCompiler/LimitExpressionTests.cs
+++ b/test/EntityFramework/FunctionalTests/PlanCompiler/LimitExpressionTests.cs
@@ -1592,7 +1592,7 @@ namespace PlanCompilerTests
     FROM  [dbo].[ArubaOwners] AS [Extent1]
     LEFT OUTER JOIN [dbo].[ArubaRuns] AS [Extent2] ON [Extent1].[Id] = [Extent2].[Id]
     WHERE N'Diego' = [Extent1].[FirstName]
-    ORDER BY [Extent1].[LastName] ASC
+    ORDER BY row_number() OVER (ORDER BY [Extent1].[LastName] ASC)
     OFFSET 2 ROWS FETCH NEXT 1 ROWS ONLY ";
 
             Limit_SimpleModel_OrderBy_Skip_FirstOrDefault_expectedSql =
@@ -1608,7 +1608,7 @@ namespace PlanCompilerTests
     FROM  [dbo].[ArubaOwners] AS [Extent1]
     LEFT OUTER JOIN [dbo].[ArubaRuns] AS [Extent2] ON [Extent1].[Id] = [Extent2].[Id]
     WHERE N'Diego' = [Extent1].[FirstName]
-    ORDER BY [Extent1].[LastName] ASC
+    ORDER BY row_number() OVER (ORDER BY [Extent1].[LastName] ASC)
     OFFSET 2 ROWS FETCH NEXT 1 ROWS ONLY ";
 
             Limit_SimpleModel_OrderBy_Skip_Single_expectedSql =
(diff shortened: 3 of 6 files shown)


maumar:
@anpete @divega

src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
```diff
@@ -2639,6 +2648,38 @@ public override ISqlFragment Visit(DbSkipExpression e)
             return result;
         }
 
+        /// <summary>
```
divega: I think we only use actual XML doc comments in public surface, so these should be `//` instead of `///`.

maumar: sounds good

maumar:
new version up

src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
```diff
@@ -2639,6 +2648,36 @@ public override ISqlFragment Visit(DbSkipExpression e)
             return result;
         }
 
+        // We need to order by row_number unless ordering list of entities and we are sorting based on entity key.
+        // In case of composite key, all parts of the key must be used.
+        private bool NeedsOrderByRowNumber(IList<DbSortClause> sortClauses)
+        {
+            var propertyExpressions = sortClauses
+                .Where(c => c.Expression is DbPropertyExpression)
```
divega: Nit: `OfType<DbPropertyExpression>()` works here?

divega: Ah no. You would need to project first.

src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
```diff
@@ -2639,6 +2648,36 @@ public override ISqlFragment Visit(DbSkipExpression e)
...
+                .Select(c => (DbPropertyExpression)c.Expression)
+                .ToList();
+
+            var propertyInstanceTypes = propertyExpressions
+                .Select(c => c.Instance.ResultType.EdmType)
+                .ToList();
+
+            var distinctEntityTypes = propertyInstanceTypes.OfType<EntityType>().Distinct().ToList();
+            if (distinctEntityTypes.Count != 1)
```
divega: Did you test this with self joins?

src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
```diff
@@ -2639,6 +2648,36 @@ public override ISqlFragment Visit(DbSkipExpression e)
...
+            {
+                return true;
+            }
+
+            foreach (var keyProperty in distinctEntityTypes.First().KeyProperties)
+            {
+                if (propertyExpressions.All(e => e.Property != keyProperty))
+                {
+                    return true;
```
divega: I wonder not emitting row_number() if things are ordered on the keys is worth the effort. Looks a bit involved. Thoughts?

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -132,6 +133,16 @@ public static bool TruncateDecimalsToScale
         }
 
         /// <summary>
+        /// Set this flag to false to prevent ordering using a row_number() function for queries with OFFSET operation, and order by columns directly instead.
+        /// </summary>
+        /// <remarks>Using row_number() can reduce ambiguity of the ordering result for some cases.</remarks>
+        public static bool UseOrderByRowNumberForQueriesWithOffset
```
maumar: UseLegacyOrderingForQueriesWithOffset? not sure if I like it better tho, any other suggestions for name?

test/EntityFramework/FunctionalTests/PlanCompiler/LimitExpressionTests.cs
```diff
@@ -834,6 +836,42 @@ orderby building.Address.ZipCode
...
         [Fact]
+        public void Limit_ComplexModel_OrderBy_Skip_without_rownumber()
+        {
+            var log = new StringWriter();
+            using (var context = new AdvancedPatternsMasterContext())
+            {
+                var previousFlag = SqlProviderServices.UseOrderByRowNumberForQueriesWithOffset;
+                try
+                {
+                    SqlProviderServices.UseOrderByRowNumberForQueriesWithOffset = false;
```
maumar: this test might not be safe if the queries are run in parallel

maumar: also, the flag is not part of the query cache key, so if one runs a query with flag set to true (default), then changes the flag to false and rerun the query, we will issue the same query - are we ok with this?

maumar:
@divega new version up

src/EntityFramework.SqlServer/SqlGen/SqlGenerator.cs
```diff
@@ -355,7 +358,7 @@ internal SqlGenerator(SqlVersion sqlVersion)
         // <param name="commandType"> CommandType for generated command. </param>
         // <returns> The string representing the SQL to be executed. </returns>
         internal static string GenerateSql(
-            DbCommandTree tree, SqlVersion sqlVersion, out List<SqlParameter> parameters, out CommandType commandType,
+            DbCommandTree tree, SqlVersion sqlVersion, bool useOrderByRowNumberForQueriesWithOffset, out List<SqlParameter> parameters, out CommandType commandType,
```
divega: Use the static class. You won't nee to flow it.

src/EntityFramework.SqlServer/SqlProviderServices.cs
divega: What about `SortyByRowNumberInOffsetQueries`?

divega: cc @rowanmiller @ajcvickers @anpete I can't think of anything good for the name of this flag.

ErikEJ: UseRowNumberOrderingInSkipTakeQueries
UseRowNumberOrderingInOffsetQueries

test/EntityFramework/FunctionalTests/PlanCompiler/LimitExpressionTests.cs
divega: Re query caching we could consider setting a separate "locking" internal flag the first time the value of the flag is used. That way you should still be able to set it very early before you start issuing queries but it would throw if you change it later. 

That said, I am not too concerned because this is very advanced.

cc @ajcvickers

divega: Re the brittle test. I would prefer not to add it. As long as all existing tests and new tests pass with the new behavior I think it is fine.

src/EntityFramework.SqlServer/SqlProviderServices.cs
maumar: UseRowNumberOrderingInOffsetQueries it is

maumar:
new version up

src/EntityFramework.SqlServer/SqlProviderServices.cs
```diff
@@ -132,6 +133,16 @@ public static bool TruncateDecimalsToScale
         }
 
         /// <summary>
+        /// Set this flag to false to prevent ordering using a row_number() function for queries with OFFSET operation, and order by columns directly instead.
```
divega: Consider switching to the same organization I used in https://github.com/aspnet/EntityFramework6/pull/182/files#diff-4b07700a7952daedde75b9c4de664a41L134. It would probably go along the lines of:

```
<summary>
Gets or sets a value indicating whether the ROW_NUMBER() function is used in sort expression passed to the ORDER BY clause when OFFSET is present in query. The default value of <c>true</c> is recommended to obtain query results that are stable for paging operations. The value of <c>false</c> can be used for compatibility with previous versions of EF and will cause the sort expression to be passed unmodified to the ORDER BY clause, which can lead to unstable results if the ordering is ambiguous. 
</summary>
<remarks>This flag only applies to SQL Server 2012 or later. This flag does not affect queries that have already been translated to SQL and cached, therefore applications that need to set the value to <c>false</c> for compatibility should do so before executing any queries.</remarks>
```
cc @rowanmiller @ajcvickers

divega: If we get good results in the beta I think we can revisit removing this flag altogether.

maumar:
Fixed in 8caee28305de907a23f766586e4a358299c3633c

Issue #198, opened 2017-02-09
In the CSharp and VisualBasic MigrationCodeGenerators strings are not escaped correctly


CZEMacLeod:
I encountered this issue after creating a custom annotation that serialized itself using JSON as the string.
The migration generators do not escape the quotes (for VB or C#) and, in the case of C# do not escape \'s in the string either.
The function summary for the Quote function mentions escaping appropriately, but currently does not.
I think this also applies if e.g. the default value for a field were to include a quote mark.
VisualBasicMigrationCodeGenerator.cs
```
        /// <summary>
        /// Quotes an identifier using appropriate escaping to allow it to be stored in a string.
        /// </summary>
        /// <param name="identifier"> The identifier to be quoted. </param>
        /// <returns> The quoted identifier. </returns>
        protected virtual string Quote(string identifier)
        {
            return "\"" + identifier.Replace("\"","\"\"") + "\"";
        }
```
CSharpMigrationCodeGenerator.cs
```
        /// <summary>
        /// Quotes an identifier using appropriate escaping to allow it to be stored in a string.
        /// </summary>
        /// <param name="identifier"> The identifier to be quoted. </param>
        /// <returns> The quoted identifier. </returns>
        protected virtual string Quote(string identifier)
        {
            return "\"" + identifier.Replace("\\","\\\\").Replace("\"","\\\"") + "\"";
        }
```

CZEMacLeod:
I am not sure about the C# version of the function I suggested above - but did see this function on Stackoverflow which might be an even better solution:
[http://stackoverflow.com/questions/323640/can-i-convert-a-c-sharp-string-value-to-an-escaped-string-literal](http://stackoverflow.com/questions/323640/can-i-convert-a-c-sharp-string-value-to-an-escaped-string-literal

```
        /// <summary>
        /// Quotes an identifier using appropriate escaping to allow it to be stored in a string.
        /// </summary>
        /// <param name="identifier"> The identifier to be quoted. </param>
        /// <returns> The quoted identifier. </returns>
        protected virtual string Quote(string identifier)
        {
            using (var writer = new StringWriter())
            {
                using (var provider = CodeDom.Compiler.CodeDomProvider.CreateProvider("CSharp"))
                {
                    provider.GenerateCodeFromExpression(new CodeDom.CodePrimitiveExpression(identifier), writer, null);
                    return writer.ToString();
                }
            }
        }
```

rowanmiller:
[...]

PR #202, opened 2017-02-14
Fixes bad escaping of String values in generated migration code

CZEMacLeod:
To address issue #198
String values containing quotes or other language specific strings are not escaped properly when generating migration code.
This change replaces the escaping code with System.CodeDom.Compiler.CodeDomProvider to allow generating correctly escaped strings.

### DIFF
--- a/src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
+++ b/src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
@@ -1536,7 +1536,14 @@ namespace System.Data.Entity.Migrations.Design
         /// <returns> The quoted identifier. </returns>
         protected virtual string Quote(string identifier)
         {
-            return "\"" + identifier + "\"";
+            using (var writer = new StringWriter())
+            {
+                using (var provider = CodeDom.Compiler.CodeDomProvider.CreateProvider("CSharp"))
+                {
+                    provider.GenerateCodeFromExpression(new CodeDom.CodePrimitiveExpression(identifier), writer, null);
+                    return writer.ToString();
+                }
+            }
         }
     }
 }
--- a/src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
+++ b/src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
@@ -1536,14 +1536,7 @@ namespace System.Data.Entity.Migrations.Design
         /// <returns> The quoted identifier. </returns>
         protected virtual string Quote(string identifier)
         {
-            using (var writer = new StringWriter())
-            {
-                using (var provider = CodeDom.Compiler.CodeDomProvider.CreateProvider("CSharp"))
-                {
-                    provider.GenerateCodeFromExpression(new CodeDom.CodePrimitiveExpression(identifier), writer, null);
-                    return writer.ToString();
-                }
-            }
+            return "\"" + identifier.Replace("\\", "\\\\").Replace("\"", "\\\"") + "\"";
         }
     }
 }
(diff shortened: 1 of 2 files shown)


rowanmiller:
We're not sure that this is the right fix. It seems like the purpose of this method is specifically to quote identifiers. Maybe the code that writes out the string values should be calling a separate method to escape them. Assigning to @bricelam to investigate.

src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
```diff
@@ -1536,7 +1536,14 @@ protected virtual string TranslateColumnType(PrimitiveTypeKind primitiveTypeKind
...
         {
-            return "\"" + identifier + "\"";
+            using (var writer = new StringWriter())
+            {
+                using (var provider = CodeDom.Compiler.CodeDomProvider.CreateProvider("CSharp"))
+                {
+                    provider.GenerateCodeFromExpression(new CodeDom.CodePrimitiveExpression(identifier), writer, null);
+                    return writer.ToString();
+                }
+            }
```
bricelam: I much prefer you're original suggestion to using CodeDOM. The following is probably sufficient. (It's what we use in EF Core)
```C#
return "\"" + identifier.Replace("\\", "\\\\").Replace("\"", "\\\"") + "\"";
```

src/EntityFramework/Migrations/Design/VisualBasicMigrationCodeGenerator.cs
```diff
@@ -1577,7 +1577,14 @@ protected virtual string TranslateColumnType(PrimitiveTypeKind primitiveTypeKind
...
         {
-            return "\"" + identifier + "\"";
+            using (var writer = new StringWriter())
+            {
+                using (var provider = CodeDom.Compiler.CodeDomProvider.CreateProvider("VisualBasic"))
+                {
+                    provider.GenerateCodeFromExpression(new CodeDom.CodePrimitiveExpression(identifier), writer, null);
+                    return writer.ToString();
+                }
+            }
```
bricelam: For VB, I think this is sufficient:
```C#
return "\"" + identifier.Replace("\"", "\"\"") + "\"";
```

bricelam:
@rowanmiller it's the right method. `identifier` is a poor parameter name, and it was probably copied from something generating SQL...

src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
rowanmiller: We discussed that in triage today too... we couldn't remember if we used CodeDOM already or not.

CZEMacLeod: Sorry for the delay - I have been using my version including CodeDOM for a while now and in general I think the CodeDOM approach works better when generating long strings etc. as it adds line breaks as required.
However, I did find an issue when generating migrations in a class library with a startup project which included the Microsoft.CodeDom.Providers.DotNetCompilerPlatform package as it resulted in not finding the correct version of the providers unless the package was also added to the class library. Although this did result in using the latest Roslyn based providers.
I am happy to switch to the 'simpler' Replace approach for the providers as suggested and will update my pull appropriately.

bricelam:
I've reverted this. We had some CI failures that may be related. I've resubmitted the changes in #232 so we can try again before the 6.3.0 release.

ErikEJ:
@bricelam  What is the release roadmap for EF 6.2 / EF 6.3 actually?

bricelam:
lol, no idea

divega:
@erikej The main issue we were waiting to resolve was the story for shipping a beta version of the EF6 Designer that could be downloaded out of band and installed on VS 2017. We have sorted that out (we won't do a beta for VS2017 but we will use the preview channel at a later point), so now it is a matter of making spending a few cycles to put everything together. There is also one bug that it is still a candidate to be fixed in the beta: #218. Hopefully next week we will be able to pick a date.

PR #216, opened 2017-03-11
[ComputedAttibute] Column with optional behavior

mmsgg:
This attribute should be enriched with optional  behavior that specified if or no refresh select SQL statement is generated.
That behavior apparently seems banal, but it is very important for further ef's flexibility.
When poco class element is decorated with [Computed] attribute, should be possible define an other optional parameter on which specified the behavior aforementioned.
This behavior just exist in EF core, why not should  exits in ef 6?


ErikEJ:
Looks like the wrong commits??

ajcvickers:
@mmsgg Please update with appropriate code. Also, how would this differ from `[DatabaseGenerated(DatabaseGeneratedOption.Computed)]`. Finally, I'm not saying we definitely would not take a PR for this, but the bar is pretty high for any new attribute classes.

PR #227, opened 2017-03-22
Fix misuses of Generate in migration code generators

bricelam:


src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
```diff
@@ -535,10 +535,9 @@ var indentString
 
                 writer.Write("@");
                 writer.WriteLine(
-                    Generate(
-                        procedureOperation
-                            .BodySql
-                            .Replace(Environment.NewLine, indentString)));
+                    Quote(procedureOperation.BodySql)
```
bricelam: Wait, this is wrong. Will update.

Issue #221, opened 2017-03-17
Configuration seed method aware of migrations applied.


JoseCMRocha:
Possibility to have the knowledge of the migrations applied during an update on the method seed of the class configuration.

Right know there is no easy way to know what migrations where applied when running an upgrade, so we can use that information to run only the seed/s for the target migration. 

The most common non-invasive solutions you see for this are creating a `DataBaseInitializer` that is useless running the upgrade from the package manager console, or creating a class from `DbMigrator` on the class `Configurator`, what in my opinion is simple wrong since the `DbMigrator `already exists and it will bring a lot of other problems with the context.

So what I'm proposing a new method override on the class `DBMigrationsConfiguration`, that has as parameter the context and a list with the id's of the migrations applied, very simple and noninvasive solution `protected virtual void Seed(TContext context, IEnumerable<string> migrationsApplied)` as a new signature that if it is not overrided by the user class it calls the previous `protected virtual void Seed(TContext context)`.

This will change the way we use the seeder\s today, making them more useful, or just stick with the way it were before, that's up to the user choice.

A simple note is that it this a nice to have would be a extension similar with the `AddOrUpdate` called `RemoveIfExist` the name is self explanatory. 

Ps. This is a feature "request" discussion, so I can decide the usefulness of this changes for others, and make pull a request of a fork already with this changes.

tinchou:
Hi Jose,

I'm currently working on a seed data proposal in EF Core: https://github.com/aspnet/EntityFramework/issues/629#issuecomment-282368487. I'm not suggesting we should do this in EF6, but anyway I'd like to know if this addresses your use case. The main idea is to move the data seeding code to the migration itself.

Thanks!

JoseCMRocha:
Hi tinchou,

I have read trough your proposal, and the its a good proposal but in a different context. In EF6 there is already a way of calling seeders on upgrade migrations from the method Seed on the class Configuration, what I was proposing here was a more simple way to have an overload of that method and instead of have a method that just receives the context, an overload to receive the context and a list with the ids of the migrations applied.

Your proposal is great but it implies a major development.

tinchou:
Thank you for taking a look! I'm implementing that proposal on EF Core, and since it's different from EF6 data seeding I want to be sure I'm covering the same and new scenarios.

smithaitufe:
@tinchou Thanks so much for your interest in seeding. I love the ideas you proposed. 

[...]

PR #229, opened 2017-03-23
Migrations configuration seeder aware of migrations applied.

JoseCMRocha:
Implementation of a signature overload in the `DbMigrationsConfiguration` so it can be called with a new parameter a list with the id's of the migrations that were applied during the migration upgrade.

The usage of this new method is done by replacing the method on the class "Configuration" generated by the migration, from `protected override void Seed(EfDbContext context)` to `protected override void Seed(EfDbContext context, IEnumerable<string> migrationsApplied)`. 

The both signatures are kept intact on the migration decorators in order to keep "backwards compatibility".

A different method was added to the `DbSetMigrationsExtensions` called RemoveIfExists that remove an entity if that entity exists following the same logic approach of the method `AddOrUpdate`.

A Discussion was previously started at [Configuration seed method aware of migrations applied. #221](https://github.com/aspnet/EntityFramework6/issues/221) .

### DIFF
--- a/src/EntityFramework/Migrations/DbMigrationsConfiguration.cs
+++ b/src/EntityFramework/Migrations/DbMigrationsConfiguration.cs
@@ -271,10 +271,11 @@ namespace System.Data.Entity.Migrations
                 _commandTimeout = value;
             }
         }
-
-        internal virtual void OnSeed(DbContext context)
+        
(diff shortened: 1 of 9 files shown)


ajcvickers:
Closing because we are not going to take this change, as described in the issue discussion.

Issue #247, opened 2017-04-17
EF6 Performance Optimization


lajones:
In Nov 2016 we added attributes to the bootstrap package, `EntityDesignBootstrapPackage`, which specified loading the package only when a `.edmx` file was selected (instead of at solution-load time). But this performance optimization work did not result in the performance improvement we had expected. After investigation it looks like the attributes that were added needed corresponding updates to the pkgdef file in that project as well in order to take effect.

lajones:
Fixed with PR #248.

PR #248, opened 2017-04-17
Perf optimization. Update pkgdef data to reflect UIContextRule autoload attributes

lajones:
Fix for #247. AutoLoad attributes must be kept in sync with the pkgdef files.

### DIFF
--- a/src/EFTools/EntityDesignBootstrapPackage/EntityDesignBootstrapPackage.csproj
+++ b/src/EFTools/EntityDesignBootstrapPackage/EntityDesignBootstrapPackage.csproj
@@ -95,7 +95,8 @@
       <DesignTime>True</DesignTime>
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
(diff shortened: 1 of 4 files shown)


ajcvickers:
@lajones Have you tested that this works? For example, maybe temporarily instrument the designer to throw or write to a file on load, and then try starting VS with the tab active or not and make sure that the loading happens at the correct time?

lajones:
@ajcvickers full Dev15 testing will require that we go through the Dev15 check-in process and get a signed module representing our VSIX and SingletonMSI - my attempts to hack around that caused immediate catastrophic failure of my hacked version of VS. But I was able to make Dev14 look like Dev15 as far as this is concerned - and there I was able to test. Results of the test were good i.e. installing the pkgdef with the new registry entries did have the effect of no longer loading  our package at sln load. Instead, as expected, package load happens whenever a .edmx file is selected.

Can you sign off now?

lajones:
Checked in with fb1af12da (on release-6.2.0-beta1 branch) and e207f07c7 (on master).

Issue #250, opened 2017-04-21
Excessive ToList calls


Backs:
Solution contains a lot of ToList calls, sometimes ther are excessive and leadt to unnecessary memory traffic and perfomance problems.
In some places it's better to use more generic types like IEnumerable<T>

ajcvickers:
@Backs I'm sure there are places where ToList calls are unnecessary, but usually these are there by design to protect the original collection, often because there is a possibility that it will be modified while it is being iterated. For us to take any PR on this we would have to be confident that in every place the ToList is removed there will not be a regression in functionality. Running all the test would be the minimum bar for this, but because of the history of the codebase we don't have good coverage for some areas of the stack, and in those places unless it is very clear that the change is safe we would likely err on the side of caution.

Backs:
@ajcvickers I checked many places - lot of them use some sort of IReadOnlyCollection, so here, I'm sure, ToList is unnecessary. Sometimes it's list.ToList() calls. Maybe, it's better to split this issue to small tasks like "remove ToList in method A". What do you think?

ajcvickers:
@Backs We discussed this and decided that we will consider PRs to remove ToList but only in places where:
* It can be shown with confidence that there is no functional impact
* It is in a performance critical code path, either in terms of pure performance, or in terms of creating a lot of garbage.

So if you want to pursue this, then smaller PRs with justifications for the points above would probably be the way to go.

Backs:
@ajcvickers ok, I'll split PR to small isolated parts

PR #252, opened 2017-04-22
#250 - modify MetadataCollection.AddRange method to accept IEnumerable

Backs:
MetadataCollection class has method AddRange that accepts List. Due to internal implementation, this method just iterates through list and adds every item to internal collection _metadataList and _caseSensitiveDictionary.
It looks correct to change signature of the method AddRange to accept IEnumerable, like List.AddRange. Thereby we can change signature of method MetadataItem.AddMetadataProperties to accept IEnumerable too, and remove ToList when we call AddMetadataProperties.

### DIFF
--- a/src/EntityFramework/Core/Metadata/Edm/AssociationEndMember.cs
+++ b/src/EntityFramework/Core/Metadata/Edm/AssociationEndMember.cs
@@ -88,7 +88,7 @@ namespace System.Data.Entity.Core.Metadata.Edm
 
             if (metadataProperties != null)
             {
-                instance.AddMetadataProperties(metadataProperties.ToList());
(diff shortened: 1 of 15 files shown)


ajcvickers:
@Backs Are you sure that there is no chance that IEnumerable passed in is not ever backed by a collection that could be modified by the AddInternal method? If so, this change seems fine. If not, this could cause the collection to be modified while it is being enumerated.

Backs:
@ajcvickers I need some time to restore the issue, becase it was quite a long time ago. I'll reply later.

Backs:
@ajcvickers sorry for the late answer.
I checked usage of AddRange method, as I see, it's safe.

ajcvickers:
@Backs Thanks for checking.

PR #255, opened 2017-04-25
Add or update and delete operation for data manipulation as part of the migration.

JoseCMRocha:
Implementation of operations AddOrUpdate and Delete in order to be able to upsert and delete data on tables while migrating. 
This pull request comes form a another issue [#221](https://github.com/aspnet/EntityFramework6/issues/221) where a discussion has placed and a previous implementation was discarded and resulting in this new implementation has originally suggested by @tinchou.

The work done in this pull request allows to have data manipulation as part of the migration. The operations are not bound to the model CLR types, so previous migrations will not break. 
This means that the data manipulation can be added to the migrations and implemented in SQL. 

The work done in this pull request is based in the work of @tinchou  for EF Core described in: [ aspnet/EntityFramework#629 (comment)](https://github.com/aspnet/EntityFramework/issues/629#issuecomment-282368487)

Examples of usage:

AddOrUpdate with identifier columns
```
// Migration
AddOrUpdate("foo.boo", new[] { "cA" }, new[] { "cA", "cB", "cC"}, new object[] { "vA", "vB", 1 });
// SQL generated
"IF object_id('[foo].[boo]') IS NOT NULL
 IF EXISTS (SELECT 1 FROM [foo].[boo] WHERE [cA] = 'vA') UPDATE [foo].[boo] SET [cA] = 'vA', [cB] = 'vB', [cC] = 1 WHERE [cA] = 'vA' ELSE
 INSERT INTO [foo].[boo]([cA], [cB], [cC]) VALUES ('vA', 'vB', 1)"
```

AddOrUpdate with no identifier columns
```
// Migration
AddOrUpdate("foo.boo", new[] { "cA", "cB", "cC"}, new object[] { "vA", "vB", 1 });
// SQL generated
"IF object_id('[foo].[boo]') IS NOT NULL
 INSERT INTO [foo].[boo]([cA], [cB], [cC]) VALUES ('vA', 'vB', 1)"
```

Delete
```
// Migration
Delete("foo.boo",new[] { "cA", "cB", "cC" }, new [] { new object[] { "vA1" , "vA2"}, new object[] { 2 },  new object[] { } });
// SQL generated
"IF object_id('[foo].[boo]') IS NOT NULL
 DELETE FROM [foo].[boo] WHERE ([cA] = 'vA1' OR [cA] = 'vA2') AND ([cB] = 2)"
```

### DIFF
--- a/src/EntityFramework.SqlServer/SqlServerMigrationSqlGenerator.cs
+++ b/src/EntityFramework.SqlServer/SqlServerMigrationSqlGenerator.cs
@@ -829,9 +829,9 @@ namespace System.Data.Entity.SqlServer
         {
             Check.NotNull(addOrUpdateOperation, "addOrUpdateOperation");
 
-            var columns = addOrUpdateOperation.Columns.ToList();
+            var columns = addOrUpdateOperation.Columns;
             var identifiers = columns.Intersect(addOrUpdateOperation.Identifiers).ToList();
-            var values = addOrUpdateOperation.Values.ToList();
+            var values = addOrUpdateOperation.Values;
 
             using (var writer = Writer())
             {
--- a/src/EntityFramework.SqlServer/SqlServerMigrationSqlGenerator.cs
+++ b/src/EntityFramework.SqlServer/SqlServerMigrationSqlGenerator.cs
@@ -829,26 +829,28 @@ namespace System.Data.Entity.SqlServer
         {
             Check.NotNull(addOrUpdateOperation, "addOrUpdateOperation");
 
+            var columns = addOrUpdateOperation.Columns.ToList();
+            var identifiers = columns.Intersect(addOrUpdateOperation.Identifiers).ToList();
+            var values = addOrUpdateOperation.Values.ToList();
+
             using (var writer = Writer())
             {
                 writer.Write("IF object_id('");
                 writer.Write(Name(addOrUpdateOperation.Table));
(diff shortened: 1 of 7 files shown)


ajcvickers:
@bricelam Take a look at this when you can--we won't include before 6.3 at this point.

ajcvickers:
@JoseCMRocha Bringing EF Core style seeding still seems feasible to me. However, we worked a lot on the Core experience before shipping it, and I think a lot of the changes made should also be made here before merging.  A good starting point is [MigrationBuilder](https://github.com/aspnet/EntityFrameworkCore/blob/release/2.2/src/EFCore.Relational/Migrations/MigrationBuilder.cs). It would also be good to bring over any relevant tests from Core.

/cc @divega

JoseCMRocha:
@ajcvickers when we talk about shipping more experience from core, we are talking about bringing more features from there or rework the [DbMigration](https://github.com/aspnet/EntityFramework6/blob/master/src/EntityFramework/Migrations/DbMigration.cs) from EF6? 
In the case that is to bring more features, can we define a plan, try to identify which features should be brought from core?
About the tests it is always a good idea.

ajcvickers:
@JoseCMRocha I'm talking specifically about the data motion methods in this PR, only the updated versions from where they were further developed in EF Core.

ajcvickers:

Issue #243, opened 2017-04-11
EF Tools 6.1.3 Update Model Issue


stephanebo21:
Sorry to post again my issue in reference of issue #109.

Sorry for my late answer, but I needed time to install a new clean machine to reproduce this case.
I have reproduced the issue and can give you details.
Before when the solution worked, after updating my EF Model I could encounter this kind of warning:

> warning 6002: The table/view 'myView' does not have a primary key defined. The key has been inferred and the definition was created as a read-only table/view.

This issue was resolved by the development of a custom tool which modify xml model after the update and everything worked fine.

Now, since I am trying to work with Visual Studio 2015, EF 6.1.3, when I try to add the same view I have this error:

> Error 13101: Key part ‘JOB_NAME’ for type ‘JOBS’ is not valid. All parts of the key must be non-nullable.

There is no way to modify xml because my view is not added to the Model Store.
I am pretty sure at 100% that this problem is specific to users which use EF and Oracle. All references I found on the internet are from people who have the same development context.

This is a critical risk for us because we have no way to upgrade our application to new versions…

Seems like mySQL users have some issues similar from exactly the same version of Visual Studio and Entity Framework:
https://social.msdn.microsoft.com/Forums/es-ES/544bfe56-b6e3-478e-993e-b549f9cce712/error-13101-key-part-nombre-for-type-recursos-is-not-valid-all-parts-of-the-key-must-be?forum=aspnetmvces

PS: I Have no problem with SQL Server views.

Regards
Stéphane

lajones:
@stephanebo21 - so, just to be clear, you are saying this Entity no longer shows up _even if you open the .EDMX file in an XML editor and search for `JOBS`_? It might be there just commented out. (If it is just commented out it would not show up at all when you open the file in the EDMX designer, but you might be able to solve the problem by manually editing the XML to declare your own key).

divega:
@stephanebo21 I have updated this new bug report with the additional information from your comment at https://github.com/aspnet/EntityFramework6/issues/109#issuecomment-273514897. For future reference, it should always be possible to re-open an issue if you are the original poster. You should also be able to edit your own comments to add any missing information. 

Regarding the issue itself, the two messages reveal that something has changed in the metadata we get about the view from the database between the two versions: 

> Warning 6002: The table/view 'myView' does not have a primary key defined. The key has been inferred and the definition was created as a read-only table/view.

[...]

PR #259, opened 2017-04-28
Nullable Keys are now being allowed.

lajones:
Fix for #243. Previously we disallowed nullable key properties. This fix allows them.

### DIFF
--- a/src/EntityFramework/Core/Metadata/Edm/EdmModelRuleSet.cs
+++ b/src/EntityFramework/Core/Metadata/Edm/EdmModelRuleSet.cs
@@ -63,7 +63,6 @@ namespace System.Data.Entity.Core.Metadata.Edm
             AddRule(EdmModelSemanticValidationRules.EdmAssociationSet_DuplicateEndName);
             AddRule(EdmModelSemanticValidationRules.EdmEntityType_EntityKeyMustBeScalar);
             AddRule(EdmModelSemanticValidationRules.EdmEntityType_DuplicatePropertyNameSpecifiedInEntityKey);
-            AddRule(EdmModelSemanticValidationRules.EdmEntityType_InvalidKeyNullablePart);
             AddRule(EdmModelSemanticValidationRules.EdmEntityType_InvalidKeyKeyDefinedInBaseClass);
             AddRule(EdmModelSemanticValidationRules.EdmEntityType_KeyMissingOnEntityType);
             AddRule(EdmModelSemanticValidationRules.EdmEntityType_InvalidMemberNameMatchesTypeName);
--- a/src/EntityFramework/Core/Metadata/Edm/EdmModelSemanticValidationRules.cs
+++ b/src/EntityFramework/Core/Metadata/Edm/EdmModelSemanticValidationRules.cs
@@ -310,26 +310,6 @@ namespace System.Data.Entity.Core.Metadata.Edm
                         }
                     });
 
-        internal static readonly EdmModelValidationRule<EntityType> EdmEntityType_InvalidKeyNullablePart =
-            new EdmModelValidationRule<EntityType>(
-                (context, edmEntityType) =>
-                    {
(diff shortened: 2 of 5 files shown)


divega:
I am surprised that it was enough to remove the validation rule. Don't we also have to change some code to always set key members as non-nullable when we are processing the schema discovery queries?

lajones:
@divega That sounds contrary to the approach I thought @ajcvickers was suggesting. I thought the approach was to allow things to be nullable where possible (knowing that that might cause problems at runtime if there actually is a null in the DB). See [this comment](https://github.com/aspnet/EntityFramework6/issues/243#issuecomment-294220463).

However that comment does also mention a warning from the scaffolding tool - I should add that. But I'd like to get this in so I can "re-freeze" the runtime.

divega:
@lajones I think I understand how you read that comment, however it clearly says we need to change the designer :smile:

ajcvickers:
@lajones I'm sorry my comment was not clear enough. Let me try to be more specific.

The problem, as I understand it, is:
* Nullable columns are allowed for primary keys in some database engines
* If a column in a key comes back as nullable, then some piece of the reverse engineering code says that that is invalid because EF6 doesn't allow nullable key properties
* However, if the user was not using reverse engineering and setup their model using non-nullable properties*, then EF6 would not complain (as long as there aren't actually any null key values).

Therefore the solution we want to get to is for reverse engineering to generate the model that could have been setup by hand using non-nullable properties*. (Ideally it should generate a warning when doing this indicating that the nullability of the key columns is being ignored and that if there are null keys values, then this will not work at runtime because EF doesn't support null key values. But I could live without the warning if this is a lot of work.)

So, my comment was meant to convey that the gist of all this is to stop rejecting the columns/keys in reverse engineering _and then also ensure that the generated model actually works_, doing whatever property coercion is needed, etc. Sorry I wasn't clear about this. 

* When I say non-nullable properties, I'm not sure if the CLR type needs to be non-nullable, if the c-space type needs to be non-nullable, or if only the s-space type needs to be non-nullable. This is something that I would expect you to determine as part of the fix--probably just by editing an EDMX to see what happens in the different cases. Beyond that, even if the CLR types and/or the c-space type can nullable, given that null values are not allowed I would probably err on the side of making the property non-nullable at all levels, since I think that better matches the semantics of what is happening, and null doesn't have special meaning like it does in EF Core. But hopefully you can look at the code, try a few things, and make a recommendation on this. Probably also worth asking @divega's opinion.

divega:
@lajones @ajcvickers my mental model of how it currently works is mostly the same:

1. Reverse engineering executes a schema discovery query and gets back a table with columns and key metadata. For whatever reason (possibly a limitation of the provider, the database or the schema discovery queries) the metadata about the keys specifies that they are nullable S-space types. 
2. Reverse engineering constructs a model with the table/column/key metadata it obtained. While doing so it reflects the fact that the keys are reported as nullable into the model.
3. Then the  pass this to the model through semantic validation. The validation rule from the runtime throws because EF does not accept nullable key members. 

Here is what I think we want to happen after the fix:

1. Reverse engineering executes a schema discovery query and gets back a table with columns and key metadata. For whatever reason (possibly a limitation of the provider, the database or the schema discovery queries) the metadata about the keys specifies that they are nullable S-space types. 
2. Reverse engineering constructs a model with the table/column/key metadata it obtained. While doing so it **ignores the fact that the keys are reported as nullable** and makes them always non-nullable in the model.
3. Then the  pass this to the model through semantic validation. The validation rule from the runtime **passes** because all keys are non-nullable. 

I agree with @ajcvickers that a warning would be nice if in step (2) we find that the keys were reported as nullable, but I believe it is very low priority in my mind because the likelihood of a relational database reporting that PKs are nullable because they are really nullable is extremely low.

lajones:
Closing this PR as it doesn't address the issue the way that was meant. Will open a new PR addressing this in the designer.

PR #262, opened 2017-05-02
Nullable PK columns now result in non-nullable PK properties.

lajones:
Fix for #243. This fix coerces the generated S-side property (and thus the generated C-side property as well) to be non-nullable if the underlying column was nullable but was also marked as part of the primary key. This helps when reverse engineering some Views on Oracle where they can report nullable primary keys.

Notes: I tried coercing just the C-side property, but Model Validation checks S-side properties as well - so it would not have solved the customer's issue.

For tests: Where possible I updated the test so that it would just do the same thing as before; sometimes that required more constrained input data. Note: changing the nullability of columns can change how we calculate the multiplicity of the association ends.  In one case I had to split the existing test into 2 separate tests: PK-to-PK and PK-to-FK associations. Previously only PK-to-PK association was checked - some of the scenarios it checked are no longer possible. So I removed those but added the new PK-to-FK test (I checked that the new test would have gotten the same results before this change).


src/EFTools/EntityDesignerVersioningFacade/Resources.VersioningFacade.resx
```diff
@@ -207,4 +207,7 @@
   <data name="EF6Providers_NotDbProviderServices" xml:space="preserve">
     <value>The 'Instance' member of the Entity Framework provider type '{0}' did not return an object that inherits from 'System.Data.Entity.Core.Common.DbProviderServices'. Entity Framework providers must extend from this class and the 'Instance' member must return the Singleton instance of the provider.</value>
   </data>
+  <data name="CoercingNullablePrimaryKeyPropertyToNonNullable" xml:space="preserve">
+    <value>Primary key column {0} on table {1} is nullable. Entity Framework does not allow non-null primary keys so it will coerce the matching primary key property {2} to be non-nullable.</value>
```
divega: Minimum change:

> ... Entity Framework does not allow **null** primary keys...

Suggested message: 

> The Entity Framework provider in use reported the primary key column {0} on table {1} to be nullable. Entity Framework does not allow nullable primary keys, therefore any null values stored in this column could cause errors at run time.

Note that:
- Rather than say the provider in use we could use the invariant name to identify it (I believe it is available as a field in the class)
- Talking about the property and even the name of the property is redundant in this context.

lajones: Good catch on the "non-null". Agree that `_providerInvariantName` is available if we want to use it in the message. @ajcvickers what do you think? I included the name of the property as it can be different from the name of the column due to clashes and it's the property that users see in the designer (but that is an unlikely scenario).

ajcvickers: I don't feel strongly one way or the other.

ajcvickers: @divega With regard to creating new strings, let's just proceed as normal for now.

lajones: I like the comment "any null values stored in this column could cause errors at run time" as that's useful for the user to know, but I like the message to say what it is doing about the situation. @divega's suggested message does not say that we are making the generated properties non-nullable which I think we should mention.

ajcvickers: I'm fine with all the suggestions made, even just "Entity Framework does not allow null primary keys." I don't think the exact message matters much.

divega: > suggested message does not say that we are making the generated properties 

Let's try this:

> The Entity Framework provider in use reported the primary key column {0} on table {1} to be nullable. Entity Framework does not allow nullable primary keys, therefore it will assume the column is not nullable and any null values stored in this column will cause errors at run time.

lajones: LGTM.

lajones:
Checked in with commit 2a8432fb3 (release-6.2.0-beta1).

Issue #152, opened 2016-12-14
PowerTools: get it working on VS2015 and VS2017


lajones:
PowerTools will need to be updated to work on VS2017. This will require access to the VS2017 SDK which should be available soon, and will require updating the vsix to V3 - again details on this should be available soon.

ErikEJ:
@lajones Any particular reason why the PowerTools solution is currently targeting VS 2013? In other words - can I update to VS 2015?

julielerman:
Also, my PR to get the VSIX fixed up for VS2015 was accepted but is not yet on VS Gallery. It's clearly needed as my blog post for doing it manually has now had almost 70K views ..still going at nearly 200 new views every day.

lajones:
@ErikEJ - no objection from me - I think it was just the last time this was touched.
@julielerman - @rowanmiller is usually responsible for releasing to VS Gallery - but he's on vacation. @diego - can you issue a release?

lajones:
See #154.

divega:
To summarize the current plan to ship PowerTools, we will:

1. Remove the reverse engineering feature (@erikej has volunteered to send a PR)
2. Do some smoke testing on the other features, in particular on VS 2015 U3 and VS 2017
3. Publish the VSIX to the gallery
4. In the release notes we note that customers can use the reverse engineering feature in the EF Tools or the Reverse POCO template for more flexibility.

ErikEJ:
@divega  Good plan!

lajones:
Note from #264 and commit 4fe5c3c. Had to update the build to only work on VS2017. This is required so that the build will generate manifest.json and catalog.json within the resulting VSIX which are required for the VSIX to install on VS2017 and are ignored on VS2015 & VS2012 so it will install and run on those too.

ErikEJ:
@lajones "Had to" -  This is actually not required, unless you are unable/unwilling to build with msbuild14.

ErikEJ:
@divega Close this? https://marketplace.visualstudio.com/items?itemName=ErikEJ.EntityFramework6PowerToolsCommunityEdition is published

divega:
Close. We are going to recommend customers to use the Power Tools Community edition moving forward.

PR #264, opened 2017-05-04
Updating PowerTools to only build on VS2017 

lajones:
(but still install on older versions). This is required so that the build will generate manifest.json and catalog.json within the resulting VSIX which are required for the VSIX to install on VS2017.

Fix for #152


lajones:
Note: I can revert the `#line` changes if you think it's important. But given that I'm going to be out for a few days I thought you'd like to see the changes before I leave.

lajones:
Fix checked in with commit 4fe5c3c.

PR #267, opened 2017-05-08
Possible null dereference fix

Backs:
If first condition is false, second one can cause NullReferenceException


Backs:
Sorry, wrong branch :(

PR #268, opened 2017-05-08
Possible null dereference fix

Backs:
If first condition is false, second one can cause NullReferenceException

### DIFF
--- a/src/EntityFramework/Core/Mapping/ObjectTypeMapping.cs
+++ b/src/EntityFramework/Core/Mapping/ObjectTypeMapping.cs
@@ -96,10 +96,9 @@ namespace System.Data.Entity.Core.Mapping
         {
             var memberMapping = GetMemberMap(propertyName, false /*ignoreCase*/);
 
-            if (memberMapping != null &&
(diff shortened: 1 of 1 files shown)


src/EntityFramework/Core/Mapping/ObjectTypeMapping.cs
```diff
@@ -96,10 +96,9 @@ internal ObjectPropertyMapping GetPropertyMap(String propertyName)
         {
             var memberMapping = GetMemberMap(propertyName, false /*ignoreCase*/);
 
-            if (memberMapping != null &&
-                memberMapping.MemberMappingKind == MemberMappingKind.ScalarPropertyMapping
+            if (memberMapping?.MemberMappingKind == MemberMappingKind.ScalarPropertyMapping
                 ||
-                memberMapping.MemberMappingKind == MemberMappingKind.ComplexPropertyMapping)
```
smitpatel: put conditions around `||` in brackets so we don't check for not null twice.

Backs:
@smitpatel done

Issue #20, opened 2016-07-29
Breaking Change: Query crashes when compiled with VS2015 RTM (Roslyn compiler)


rowanmiller:
The following query crashes when compiled with VS2015 RTM, but works when compiled with the old compiler in VS2013: 

```
public async Task<List<ProductRelation>> FailingQuery(SampleEnum queryOption)
        {
            using (var context = new SampleContext())
            {

                int[] ids = await Task.FromResult(new[] { 1, 2, 3, 4 });
                return await (from order in context.Set<Order>()
                              where ids.Contains(order.OrderId)
                                 && queryOption == SampleEnum.Value1
                              from pr in context.Set<ProductRelation>()
                              select pr).ToListAsync();
            }
        }
```

(Not that the query has been simplified for demonstration purposes, it obviously makes no sense in this shape but I have reduced it as much as possible).

```
Exception: 

System.NotSupportedException: LINQ to Entities does not recognize the method 'System.Data.Entity.DbSet`1[RoslynEfTest.UnitTest1+ProductRelation] Set[ProductRelation]()' method, and this method cannot be translated into a store expression.
```

I have attached a sample unit-test project which illustrates the error.

rowanmiller:
Ported from CodePlex https://entityframework.codeplex.com/workitem/2785

jcchalte:
Hello,

Is there any news about this bug ? We have a huge codebase and cannot migrate to VS2015 with this bug active. I have retrieved the solution provided in the codeplex forum and tried the code given in both VS 2013 and VS 2015. Indeed, the code does not work in VS2015.

It seems that the expression that EF is trying to convert is not the same in both environments :

In VS2015, this is as following :
`value(System.Data.Entity.Core.Objects.ObjectQuery``1[RoslynEfTest.UnitTest1+Order]).MergeAs(AppendOnly).MergeAs(AppendOnly).Where(order => (value(System.Int32[]).Contains(order.OrderId) AndAlso ([-1] == 0))).SelectMany(order => value(RoslynEfTest.UnitTest1+SampleContext).Set(), (order, pr) => pr)
`
In VS2013, the expression is :
[...]

PR #285, opened 2017-05-24
Add support for accessing Set<T> method from derived dbcontext

davidroth:
Add support for accessing Set<T> methods from derived dbcontext so that all queries work properly in linqpad.
For example the usage of `Set<Category>` in the following query fails in linqpad atm:

```csharp
var queryOption = SampleEnumGitHub20.Value1;
int[] ids = { 1, 2, 3, 4 };
return await (from product in Set<Product>()
   where ids.Contains(product.Id)
       && queryOption == SampleEnumGitHub20.Value1
    from category in Set<Category>() // 
    select category).ToListAsync();
```

Current exception:

> LINQ to Entities does not recognize the method 'System.Data.Entity.DbSet`1[SimpleModel.Category] Set[Category]()' method, and this method cannot be translated into a store expression.

This PR is related to https://github.com/aspnet/EntityFramework6/issues/20 although I do not know if this worked in VS 2013.


ajcvickers:
Merged; thanks!

Issue #290, opened 2017-05-31
powertools build error.


julianhu1979:
I am using vs2017 to build the lastest soruce code of powertools. There are some dependenced libraries missed. such as 1)Microsoft Development Environment 8.0 (Version 7.0 Object Model),
                          2)Microsoft Development Environment 8.0,
                          3)Microsoft.VisualStudio.ComponentModelHost,
                          4)Microsoft.VisualStudio.Shell.11.0,
                          5)Microsoft.VisualStudio.Shell.Design

ajcvickers:
@lajones Can you provide any help/info here?

lajones:
@julianhu1979 @ajcvickers For those errors, I would echo @ErikEJ's comment above. It looks like you're missing the VS Extension Development workload. See the selected workload in the VS Installer below.
![170612_vsextensiondevelopmentworkload](https://user-images.githubusercontent.com/1915927/27056792-6d7fe102-4f7e-11e7-9323-cc575f26f589.JPG)

Note: I tried this out myself. With the latest released version of VS (15.2) I am seeing different errors to do with not being able to load `EnvDTE` and `EnvDTE80`. So we need to update the references in PowerTools.csproj to just `Reference` those, rather than using the old `COMReference`. After that the build works as expected. We should keep this open to track that work.

ErikEJ:
I have logged: https://github.com/ErikEJ/EntityFramework6PowerTools/issues/12

ErikEJ:
Close this? https://marketplace.visualstudio.com/items?itemName=ErikEJ.EntityFramework6PowerToolsCommunityEdition is now published

divega:
Per @ErikEJ's comment, we are going to recommend customers to use the Community Edition moving forward.

PR #300, opened 2017-06-12
Use standard references for EnvDTE and EnvDTE80

lajones:
Addresses build error mentioned in [comment](https://github.com/aspnet/EntityFramework6/issues/290#issuecomment-307945363) about how to reference `EnvDTE` and `EnvDTE80` in PowerTools project.

### DIFF
--- a/src/PowerTools/PowerTools.csproj
+++ b/src/PowerTools/PowerTools.csproj
@@ -53,6 +53,8 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="EnvDTE" />
+    <Reference Include="EnvDTE80" />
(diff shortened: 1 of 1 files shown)


src/PowerTools/PowerTools.csproj
```diff
@@ -53,6 +53,8 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="EnvDTE" />
```
ErikEJ: I actually have this: https://github.com/ErikEJ/SqlCeToolbox/blob/master/src/GUI/SqlCe35Toolbox/VSToolbox.csproj#L83

lajones:
@ErikEJ  You're right - that looks better - assuming it works on my box I'll update to that.

Issue #316, opened 2017-07-13
Changing WithRequiredPrincipal to WithRequiredDependent will not drop and recreate foreign key in migration if of type Guid


lauxjpn:
If you want to change the owner table of a foreign key that is of type `Guid` in a one-to-one relationship by switching from `WithRequiredPrincipal()` to `WithRequiredDependent()` within your `EntityTypeConfiguration`, the resulting migration script will drop any related indices, but will keep the old foreign key (which is now out of sync with the configuration because it is owned by the wrong table).

As a workaround, you can first create a migration that drops the foreign key by using `Ignore()` on the navigation property. After applying the migration, you remove the `Ignore()` call, make your relationship changes and create and apply a second migration.

The foreign key constraint is dropped and recreated correctly if the foreign/primary key is of type Int32.

EntityFramework Version 6.1.3

ajcvickers:
**EF Team Triage:** This issue is not something that our team is planning to address in the EF6.x code base.  This does not mean that we would not consider a community contribution to address this issue.

Moving forwards, our team will be fixing bugs, implementing small improvements, and accepting community contributions to the EF6.x code base. Larger feature work and innovation will happen in the EF Core code base (https://github.com/aspnet/EntityFramework).

**Closing an issue in the EF6.x project does not exclude us addressing it in EF Core. In fact, a number of popular feature requests for EF have already been implemented in EF Core (alternate keys, batching in SaveChanges, etc.).**

*BTW this is a canned response and may have info or details that do not directly apply to this particular issue. While we'd like to spend the time to uniquely address every incoming issue, we get a lot traffic on the EF projects and that is not practical. To ensure we maximize the time we have to work on fixing bugs, implementing new features, etc. we use canned responses for common triage decisions.*

PR #319, opened 2017-07-14
Failing unit test to demonstrate issue #316.

lauxjpn:


ajcvickers:
Closing test PR as bug has been triaged as won't fix for EF6. A PR that also contains a fix is something we would consider taking.

Issue #302, opened 2017-06-15
Idea: componentized designer wizard 


divega:
Currently when you choose to add an "ADO.NET Entity Data Model" to a project we offer a fixed set of options: 

![image](https://user-images.githubusercontent.com/1253414/27255659-af446270-5357-11e7-814a-86a8063bc271.png)

Considering:
 
- recent feedback indicating that some customers used to EF6 don't know how to start using EF Core 
- the existence of third party utilities that allow creating EF models such as the [@sjh37's Reverse POCO generator](https://marketplace.visualstudio.com/items?itemName=SimonHughes.EntityFrameworkReversePOCOGenerator) or @ErikEJ's  [SQLite Toolbox](https://marketplace.visualstudio.com/items?itemName=ErikEJ.SQLServerCompactSQLiteToolbox), and also commercial offerings such as https://www.llblgen.com/ and https://www.devart.com/entitydeveloper/,
- feedback about having custom items representing pre-made models, like https://data.uservoice.com/forums/72025--closed-entity-framework-core-feature-suggestions/suggestions/7576092-enable-extending-entity-framework-model-wizard-to 

we are starting to think about:

- Turning the first page of the wizard into an extensible launch pad for our own and third party extensions
- Adding default EF Core options there
- Possibly standardizing on ways to differentiate between options that 
  - Create empty models vs. models from existing databases
  - Support Entity Framework 6 or below vs. Entity Framework Core
- Possibly changing the name of "ADO.NET Entity Data Model" to something more accurate, e.g. "Entity Framework model".

ErikEJ:
@sjh37 !

xperiandri:
Could you create a new wizard in WPF by reusing some available options? This Windows Forms dialog is extremely outdated.

divega:
@xperiandri we would consider that if there is a lot of value in it, otherwise we will go with minimal changes. Could you explain what the advantages would be?

ajcvickers:
**EF Team Triage:** This issue is not something that our team is planning to address in the EF6.x code base.  This does not mean that we would not consider a community contribution to address this issue.

Moving forwards, our team will be fixing bugs, implementing small improvements, and accepting community contributions to the EF6.x code base. Larger feature work and innovation will happen in the EF Core code base (https://github.com/aspnet/EntityFramework).

[...]

PR #335, opened 2017-08-10
Add a telemetry event for the new model wizard

### LINKED ISSUE #332: Instrument EF Tools wizard for usage data
We would like to have more information about how much customers use each of the different options, e.g. code first from existing database, designer from new database, etc. to be able to make better decision on where we invest.

bricelam:
Resolves #332

### DIFF
--- a/src/EFTools/EntityDesign/VisualStudio/ModelWizard/ModelObjectItemWizard.cs
+++ b/src/EFTools/EntityDesign/VisualStudio/ModelWizard/ModelObjectItemWizard.cs
@@ -36,6 +36,10 @@ namespace Microsoft.Data.Entity.Design.VisualStudio.ModelWizard
     using IOleServiceProvider = Microsoft.VisualStudio.OLE.Interop.IServiceProvider;
     using Resources = Microsoft.Data.Entity.Design.Resources;
 
+#if VS14ORNEWER
+    using Microsoft.VisualStudio.Telemetry;
+#endif
+
     /// <summary>
     ///     Visual Studio invokes this wizard when a new item of type "ADO.NET Entity Data Model" is added
     ///     to an existing project.  This wizard is registered in the .vstemplate file item template.
@@ -298,6 +302,20 @@ namespace Microsoft.Data.Entity.Design.VisualStudio.ModelWizard
         [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling")]
         public void RunFinished()
         {
+#if VS14ORNEWER
+            TelemetryService.DefaultSession.PostEvent(
+                new TelemetryEvent("vs/entityframework/entitydatamodelwizard")
+                {
+                    Properties =
+                    {
+                        {
+                            "VS.EntityFramework.EntityDataModelWizard.GenerationOption",
+                            _modelBuilderSettings.GenerationOption
+                        }
+                    }
+                });
+#endif
+
             if (_edmxItem == null)
             {
                 if (_modelBuilderSettings.GenerationOption == ModelGenerationOption.EmptyModelCodeFirst
--- a/tools/EFToolsVsReferences.settings.targets
+++ b/tools/EFToolsVsReferences.settings.targets
@@ -11,10 +11,8 @@
     <Reference Include="Microsoft.VisualStudio.Shell.Immutable.12.0" />
     <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.14.0" />
     <Reference Include="Microsoft.VisualStudio.TextTemplating.14.0" />
-    <Reference Include="Microsoft.VisualStudio.Telemetry" />
   </ItemGroup>
   <ItemGroup Condition="$(VisualStudioVersion)=='15.0'">
     <Reference Include="Microsoft.VisualStudio.Shell.Framework" />
-    <Reference Include="Microsoft.VisualStudio.Telemetry" />
   </ItemGroup>
 </Project>
--- a/tools/EFToolsVsReferences.settings.targets
+++ b/tools/EFToolsVsReferences.settings.targets
@@ -11,8 +11,10 @@
     <Reference Include="Microsoft.VisualStudio.Shell.Immutable.12.0" />
     <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.14.0" />
     <Reference Include="Microsoft.VisualStudio.TextTemplating.14.0" />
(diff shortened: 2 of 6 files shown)


src/EFTools/EntityDesign/VisualStudio/ModelWizard/ModelObjectItemWizard.cs
```diff
@@ -298,6 +302,20 @@ private static void PopluateReplacementDictionary(Project project, Dictionary<st
...
         {
+#if VS14ORNEWER
+            TelemetryService.DefaultSession.PostEvent(
+                new TelemetryEvent("vs/entityframework/entitydatamodelwizard")
+                {
+                    Properties =
+                    {
+                        {
+                            "VS.EntityFramework.EntityDataModelWizard.GenerationOption",
+                            _modelBuilderSettings.GenerationOption
```
bricelam: We may want to split this out into two properties: IsFromDatabase & IsCodeFirst (or similar)

tools/EFToolsVsReferences.settings.targets
```diff
@@ -11,8 +11,10 @@
     <Reference Include="Microsoft.VisualStudio.Shell.Immutable.12.0" />
     <Reference Include="Microsoft.VisualStudio.TextTemplating.Interfaces.14.0" />
     <Reference Include="Microsoft.VisualStudio.TextTemplating.14.0" />
+    <Reference Include="Microsoft.VisualStudio.Telemetry" />
```
bricelam: We could also use the `Microsoft.VisualStudio.Telemetry` NuGet package, but referencing the private assembly directly seemed easier.

src/EFTools/EntityDesign/VisualStudio/ModelWizard/ModelObjectItemWizard.cs
```diff
@@ -298,6 +302,20 @@ private static void PopluateReplacementDictionary(Project project, Dictionary<st
         [SuppressMessage("Microsoft.Maintainability", "CA1506:AvoidExcessiveClassCoupling")]
         public void RunFinished()
         {
+#if VS14ORNEWER
+            TelemetryService.DefaultSession.PostEvent(
+                new TelemetryEvent("vs/entityframework/entitydatamodelwizard")
```
bricelam: Not really sure what to use here. I tried to follow [the guidelines](https://wiki.vsdata.io/instrumentation_rules_naming_guidance).

tools/EFToolsVsReferences.settings.targets
divega: Should we have an itemgroup for VisualStudioVersion greater than 14? The current references there seem to be non overlapping.

src/EFTools/EntityDesign/VisualStudio/ModelWizard/ModelObjectItemWizard.cs
divega: How does this get serialized? An int, an enum value or the or the ToString() of the  `ModelGenerationOption` member?

Are there restrictions on what it can be? 

It seems that the string could be more future-proof if e.g., if we implement https://github.com/aspnet/EntityFramework6/issues/302 it could keep working the same for the existing options but new options could be added with a unique string.

bricelam: It uses `ToSTring()`.

tools/EFToolsVsReferences.settings.targets
bricelam: Using the NuGet package looks cleanest. I'll update to that. It also prevents you from using API newer than Dev14.

tools/EntityFramework.settings.targets
```diff
@@ -24,7 +24,8 @@ Project global pre-settings.
     <VersionReleasePrefix>0</VersionReleasePrefix>
     <VersionRelease>$(VersionReleasePrefix)-beta2</VersionRelease>
     <VersionReleaseName>Beta 2</VersionReleaseName>
-    <ToolingDotNetFrameworkVersion Condition="'$(VisualStudioVersion)' == '11.0' OR '$(VisualStudioVersion)' == '12.0' OR '$(VisualStudioVersion)' == '14.0'">v4.5</ToolingDotNetFrameworkVersion>
+    <ToolingDotNetFrameworkVersion Condition="'$(VisualStudioVersion)' == '11.0' OR '$(VisualStudioVersion)' == '12.0'">v4.5</ToolingDotNetFrameworkVersion>
+    <ToolingDotNetFrameworkVersion Condition="'$(VisualStudioVersion)' == '14.0'">4.5.1</ToolingDotNetFrameworkVersion>
```
bricelam: Dev14 wouldn't compile locally for me against 4.5. @lajones Do you think it's it safe to update to 4.5.1?

MagicAndre1981: > @lajones Do you think it's it safe to update to 4.5.1?

better to to 4.5.2, [4.5 and 4.5.1 are no longer supported](https://blogs.msdn.microsoft.com/dotnet/2015/12/09/support-ending-for-the-net-framework-4-4-5-and-4-5-1/) only 4.5.2

bricelam: *Running* on 4.5.1 may not be supported, but I think it's safest to *compile* against it here. On a clean install of VS 2015, they may only have 4.5.1, and I wouldn't want to prevent it from working just because it's unsupported. (My two cents.)

bricelam: I think I remember reading somewhere that even though it comes with .NET 4.6 most of it needs to run on 4.5 so it can run `devenv /setup` during install without rebooting. ...but that doesn't really apply to the EF6 Tools.

bricelam: I'm good either way. I'll let other team member weigh in.

ErikEJ: I think that relates to VS 2017 not installing .NET 4.6.2 by default

ajcvickers: Agree with @bricelam that we should compile against 4.5.1.

MagicAndre1981: a higher .net version (4.6, 4.7) replaces the older one, so VS2015 RTM comes with .net 4.6 RTM (4.6.81.0) which replaces all 4.5.x releases. It was just a small hint when I looked at this PR.

ajcvickers:
Close per #332

Issue #283, opened 2017-05-23
Revisit whether DbFunctions.Like() should be an extension method over string as it currently is


divega:
Two overloads of `DbFunctions.Like()` were added as part of the fix for #241. They are extension methods over string rather than regular static methods.

I think this may have gone under the radar in API review. In the past we have avoided creating such methods as extension methods over common types because they can be very confusing when they show up outside of the context of LINQ queries. FWIW, in EF Core `Like()` is an extension method, but over the `DbFunctions` instance in `EF.Functions`.

PR #355, opened 2017-09-11
Make DbFunctions.Like not extension methods on string

ajcvickers:
Issue #283


ajcvickers:
Signed off by @divega in triage.

Issue #195, opened 2017-02-03
Restore compabibility with providers that bypass schema discovery query execution (e.g. Oracle)


divega:
EF6 Tools 6.2 beta1 will contain an improvement that eliminates redundant parameters added for filtering in database schema discovery queries (see original fix in https://github.com/aspnet/EntityFramework6/pull/41 from 
@bengutt and later changes in https://github.com/aspnet/EntityFramework6/pull/193). 

While the main motivation for this change is to help the "Upgrade model from database" work with larger databases when using SQL Server, it should also work seamlessly with most providers. 

That said, we believe that this change could have compatibility problems with certain providers if they bypass schema discovery query execution and produce the results of those queries by other means. Such providers would likely have a hardcoded assumption on the number and position of filtering parameters which this change could break.

cincuranet:
I'm following up [this](https://blogs.msdn.microsoft.com/dotnet/2017/05/23/announcing-ef-6-2-beta-1/) CTA. 

I've tested the [FirebirdClient](https://github.com/cincuranet/FirebirdSql.Data.FirebirdClient) `6.2.0-beta1` and _Entity Framework Tools 6.2 Beta 1_ on VS2015 and everything is fine. I tried one table model as well as huge 400+ tables model.

divega:
Thank you very much @cincuranet!

IanYates:
I asked on #185, but since it's closed I thought I should probably ask here on an open & related issue.  Is EF 6.2 tooling present in the new VS 2017 Update 3  (VS 15.3) release that came out this week?

divega:
@IanYates no, we don’t have public builds of the updated EF tools that target Visual Studio 2017 yet. We plan to make the updated tools part of an upcoming update of Visual Studio. The only way you can test this right now is with a previous version of Visual Studio.

IanYates:
@divega - thanks.  I've just set up a new dev machine.  I'll grab the custom EF 6.x DLLs from my old machine's VS 2017 installation and bring them over to this new machine too.  It all worked quite well - it's a shame it's taking so long to get the EF6.2 release into VS2017.  
Of course I understand there are a million things to include and appreciate the ability to interact with the EF 6.x team here :)  
The workaround will keep on doing its thing for me for the time being.
Cheers 🥂

divega:
We have got word from Oracle that this change impacts ODP.NET. Here is the plan to mitigate this in 6.2 RTM:

We bring the old code path back alongside the new one. 

[...]

PR #363, opened 2017-09-19
WIP: Fix for #195: Suppress schema query parameter optimization for providers that are not compatible

divega:
Abstracts the change we took to avoid creating duplicate parameters that contain the same value into a `ParameterCollectionBuilder` class, so that schema discovery queries can be generated with or without the optimization depending on whether the provider supports it.

This approach avoid as much as possible duplicating the code.

Added a couple of test variants that explicitly test the parameter optimization.

TO-DO: still working on setting up a machine in which everything builds, but wanted to get this out there to get feedback.


divega:
@ajcvickers @lajones This was also supposed to be in for our release in VS. 

The only reason I didn't merge it is because I had some difficulties setting up my environment for testing, but now that @lajones has it, it should be easier. 

Can you please take care of running the tests and merge it?

divega:
@lajones I have rebased this with current state in the parent branch. Let me know if I can do anything else to help.

lajones:
Don't worry I'm on this. But I'm working on the Accessibility fixes first.

lajones:
@divega See #505.

divega:
@lajones so this PR can now be closed, correct?

lajones:
Closing in favor of #505

PR #402, opened 2017-11-21
Allow query translator to translate new Guid(x.SomeStringField)

### LINKED ISSUE #249: Query: No support for casting/converting string to GUID/UNIQUEIDENTIFIER
I like to re-open following request 
https://entityframework.codeplex.com/workitem/1865

leotsarev:
Allow query translator to translate new Guid(x.SomeStringField)  to CAST(SomeStringField AS UNIQUEIDENTIFIER)

Fixes #249 
@ajcvickers said (in automatic reply, however...) that EF team could accept this kind of community contribution. 

1. Are you actually willing to accept this kind of change at all?
2. Let's discuss what other casts could be implemented here. 
3. I found correct place to make this change, haven't I?

### DIFF
--- a/src/EntityFramework/Core/Objects/ELinq/Translator.cs
+++ b/src/EntityFramework/Core/Objects/ELinq/Translator.cs
@@ -811,6 +811,14 @@ namespace System.Data.Entity.Core.Objects.ELinq
         private sealed class NewTranslator
             : TypedTranslator<NewExpression>
         {
+            /// <summary>
+            /// List of type pairs that constructor call new XXXX(YYY yyy) could be translated to SQL CAST(yyy AS XXXXX) call
+            /// </summary>
+            private List<Tuple<Type, Type>> _castableTypes = new List<Tuple<Type, Type>>()
+            {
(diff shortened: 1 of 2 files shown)


leotsarev:
When I should expect some kind of answer? My PR is waiting for some triage meeting or something?

ajcvickers:
@leotsarev Honestly, we don't look at EF6 PRs as often as maybe we should because of resourcing issues and because we tend to have he team focus on EF6 periodically rather than constantly jumping between Core and 6. Given that it is now the holiday period and people will be out I don't expect that we will get to EF6 issues until next year at the earliest. We do appreciate the PR and we will get to it, we just have to juggle with the schedules and priorities a bit before we get to it.

leotsarev:
Hi @ajcvickers ! It's perfectly fine for me to wait. Thanks for setting my expectations. "It should wait until we have time to return for EF6 from Core" is fine for me,

leotsarev:
I'm starting to get sad :-)

leotsarev:
Thank you very much!

PR #412, opened 2017-11-27
Update ObjectStateManager to handle 1:Many relationships in O(n)

brandondahler:
# Problem
A performance issue exists when loading 1:Many relationships via its navigation property.

## Details
When accessing a 1:Many navigation property, the underlying code fetches the collection and calls `ObjectStateManager`'s `UpdateRelationships` with all of the fetched entities.  

`UpdateRelationships` then calls `TryUpdateExistingRelationships` to attempt to perform "fix-up" on existing relationships.  This is particularly helpful for 1:1 relationships that stem off of the entities that were newly loaded and needs to continue to happen; however, when the relationship is 1:Many, there is nothing to update on the relationship.

This is evidenced by the code at L1278 of ObjectStateManager.cs on the previous side of the diff:
```C#
                            case RelationshipMultiplicity.Many:
                                // do nothing because its okay for this source entity to have multiple different targets, so there is nothing for us to fixup
                                break;
```

The performance problem comes when looking at `UpdateRelationships` and `TryUpdateExistingRelationships` together, in `UpdateRelationships` in the scope of a 1:Many set of targets.  There is an outer loop of `foreach (var someTarget in targets)`, where `targets` is the set of newly loaded entities.  Inside this loop, we call `TryUpdateExistingRelationships` for each of the `targets`, but using the same source member and key.

Inside `TryUpdateExistingRelationships`, we fetch the source member and key's relationship entries and loop through them via `foreach (var relationshipEntry in manager.FindRelationshipsByKey(sourceKey))`.  For a 1:Many relationship target member, we only execute logic if when we find a source entry that matches the target key via `if (relationshipEntry.IsSameAssociationSetAndRole(associationSet, sourceMember, sourceKey))`.

Given that the source entity has N child entities (`targets`), it will have a relationship entry for each of the N child entities.  Because we first loop through the N `targets`, then we loop through the N relationship entries, we can assign the complexity of this process as O(n^2).  This means that when there are, for instance, 40,000 entities being loaded by the navigation property, we are executing the same code (`if (targetKey == relationshipEntry.RelationshipWrapper.GetOtherEntityKey(sourceKey)) { ... } else { ... }`) 1,600,000,000 times.

# Solution
In order to eliminate the O(n^2) complexity of the process for 1:Many relationships, we have to update the `TryUpdateExistingRelationships` to allow us to provide a pre-computed lookup for a source member's relationship entries such that we can find the applicable entries for a given target key.  Since we use the same source member in the outside loop of `UpdateRelationships`, we can create the lookup in O(n) time that will have a O(1) complexity for finding the associated target key.

## Assumptions
The only assumption made is that the `sourceKeyRelationshipsLazy.Value` does not need to be updated when we call `AddEntityToCollectionOrReference` and/or `manager.AddNewRelation`.  While we are technically adding to the set of relationship entries that the source member has, I don't believe that the entry will logically be used.

## Final Code-related Note
This update does not cause the code to become O(n) in all cases, only for the logical case of 1:Many relationships where the source member is the one side of the entry and the target is the Many side of the relationship.

## Maintainer Note
This change was done during the course of business for my employer.  I have submitted an updated CLA, releasing this change to the EF6 project.

Brandon Dahler

### DIFF
--- a/src/EntityFramework/Core/Common/internal/materialization/shaper.cs
+++ b/src/EntityFramework/Core/Common/internal/materialization/shaper.cs
@@ -309,7 +309,7 @@ namespace System.Data.Entity.Core.Common.Internal.Materialization
 
                 if (
                     !ObjectStateManager.TryUpdateExistingRelationships(
-                        Context, MergeOption, associationSet, sourceMember, sourceKey, sourceRelationships, wrappedEntity, targetMember, targetKey,
+                        Context, MergeOption, associationSet, sourceMember, sourceRelationships, wrappedEntity, targetMember, targetKey,
                          /*setIsLoaded*/ true, out newEntryState))
                 {
                     // Try to find a state entry for the target key
@@ -331,7 +331,6 @@ namespace System.Data.Entity.Core.Common.Internal.Materialization
                                 MergeOption,
                                 associationSet,
                                 targetMember,
-                                targetKey,
                                 targetRelationships,
                                 targetEntry.WrappedEntity,
                                 sourceMember,
--- a/src/EntityFramework/Core/Common/internal/materialization/shaper.cs
+++ b/src/EntityFramework/Core/Common/internal/materialization/shaper.cs
@@ -303,10 +303,13 @@ namespace System.Data.Entity.Core.Common.Internal.Materialization
 
                 var manager = Context.ObjectStateManager;
                 EntityState newEntryState;
-                // If there is an existing relationship entry, update it based on its current state and the MergeOption, otherwise add a new one            
+                // If there is an existing relationship entry, update it based on its current state and the MergeOption, otherwise add a new one        
+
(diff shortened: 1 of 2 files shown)


brandondahler:
To express the significance of the issue that is resolved, below is a simple snippet of code excerpted from my my project:

```C#

            using (var context = _contextFactory.Create())
            {
                var cart = context.Carts.Find(464192);


                var stopwatch = Stopwatch.StartNew();

                context.Carts
                    .Where(c => c.Id == 464192)
                    .Include(c => c.Recipients)
                    .Load();

                stopwatch.Stop();

                Console.WriteLine($"Loading took {stopwatch.Elapsed}.");
            } 
```

The Cart entity has a simple 1:Many relationship with the Cart_Recipient entity, Cart 464192 is a real instance of a production cart has 40,521 Cart_Recipient child entities.

When executing the above code against EF 6.2 RTM, the following was printed:  "Loading took 00:07:02.2073967."

When executing the above code against a custom compilation of EF 6.2 with this pull request merged on top, the following was printed:  "Loading took 00:00:04.7714773."

PR #430, opened 2017-12-26
Query compilation performance

cjpearson:
(This is more of an issue than a pull request, but I wanted to share some code along with it. It's certainly not ready to merge.)

We've been having performance issues with large LINQ queries and EntityFramework. Some large or complex queries take a very long time to compile. They're cached so we only encounter this penalty once but with some queries taking several minutes to compile it's a very noticeable impact. The best workaround we've found so far is splitting the queries into several smaller queries and combining the results. This brings the total compilation time down to several seconds.

I've tried running the compilation through a profiler to find out what's causing the slowness and made a few changes to improve the compilation speed. I haven't tested them much beyond the unit tests, but based on my test query the results are promising.

51,000 node query
Before changes: 2+ minutes
With changes: ~10 seconds
With changes and lower (10,000) MaxNodeCountForTransformations: ~1 second

They're in a very rough state currently, but I'd like to get your thoughts to see if any of them are worth pursuing further.

### Lower MaxNodeCountForTransformations

This one definitely had the biggest impact. Even with all the other changes, the compilation was still taking over ten seconds with 90% of the time spent applying these transformations. The most expensive part of the transformations were the IsNull rules. I wasn't comfortable with changing the logic of these transformations so I stuck to lower-level changes, but if there are any algorithmic changes to these transformations it would be a big help.

### Changes to VarVec to operate directly on the underlying int[]

With VarVec sizes in the thousands, a lot of the CPU time was being spent in MoveNext and Subsumes in the VarVec class. I had some success optimizing these functions by doing bit operations on the underlying ints instead of iterating through bits. In this example I've added BitVec, a modified version of the [BitArray class](https://referencesource.microsoft.com/#mscorlib/system/collections/bitarray.cs), but the same changes could also be done using reflection.

### Store the Reversed VarMap

Instead of iterating through the dictionary each time GetReverseMap is called, store a reversed version. As part of this change, VarMap implements IDictionary instead of inheriting from Dictionary.

## Other possibilities:

There were a couple other places I looked at but wasn't able to find good solutions for.

### Change SubTreeId hashing

Each call to insert averaged about a thousand calls to SubTreeId.Equals. This suggested there were a lot of hash collisions for the query. I'm not sure if there's better way to hash while making sure equivalent SubTreeIds always have the same hash.

### NodeInfo GetHashValue

A lot of time was also spent in the GetHashValue method of NodeInfo. Perhaps there may be efficient implementation of this method.

Do any of these options seem like candidates to be added to EF? Or are there any other ways to improve the query compilation step? Thanks!

### DIFF
--- a/src/EntityFramework/Core/Query/InternalTrees/VarMap.cs
+++ b/src/EntityFramework/Core/Query/InternalTrees/VarMap.cs
@@ -2,6 +2,7 @@
 
 namespace System.Data.Entity.Core.Query.InternalTrees
 {
+    using System.Collections;
     using System.Collections.Generic;
     using System.Globalization;
     using System.Text;
@@ -9,25 +10,21 @@ namespace System.Data.Entity.Core.Query.InternalTrees
     // <summary>
     // Helps map one variable to the next.
     // </summary>
-    internal class VarMap : Dictionary<Var, Var>
+    internal class VarMap : IDictionary<Var, Var>
     {
         #region public surfaces
 
+        private Dictionary<Var, Var> map;
+        private Dictionary<Var, Var> reverseMap;
+
         internal VarMap GetReverseMap()
         {
-            var reverseMap = new VarMap();
-            foreach (var kv in this)
-            {
-                Var x;
-                // On the odd chance that a var is in the varMap more than once, the first one
-                // is going to be the one we want to use, because it might be the discriminator
-                // var;
-                if (!reverseMap.TryGetValue(kv.Value, out x))
-                {
-                    reverseMap[kv.Value] = kv.Key;
-                }
-            }
-            return reverseMap;
+            return new VarMap(reverseMap, map);
+        }
+
+        public bool ContainsValue(Var value)
+        {
+            return reverseMap.ContainsKey(value);
         }
 
         public override string ToString()
@@ -35,7 +32,7 @@ namespace System.Data.Entity.Core.Query.InternalTrees
             var sb = new StringBuilder();
             var separator = string.Empty;
 
-            foreach (var v in Keys)
+            foreach (var v in map.Keys)
             {
                 sb.AppendFormat(CultureInfo.InvariantCulture, "{0}({1},{2})", separator, v.Id, this[v].Id);
                 separator = ",";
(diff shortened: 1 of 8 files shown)


divega:
@cjpearson I am sorry that we haven't been responsive on this PR or on any of the other PRs for EF6. The team has been focusing completely on EF Core releases, but we expect to be able to do a push for making some progress on EF6.3 soon.

divega:
> The most expensive part of the transformations were the IsNull rules

@cjpearson this reminds me of 'context.Configuration.UseDatabaseNullSemantics'. Does setting this flag to true affect the compilation cost for your queries?

cjpearson:
Thanks for the update. I ran some quick tests with `context.Configuration.UseDatabaseNullSemantics = true` and it seems to help performance although not to the same degree. I haven't profiled it yet.

ajcvickers:
@cjpearson Some thoughts on each of the points above:
* MaxNodeCountForTransformations: Note sure we can change this without breaking existing scenarios that work. However, making this something that can be configured by the application does not seem unreasonable.
* The VarVec changes look promising to me. How confident are you that the functionality is not changed? Would more tests help improve that confidence?
* The SubTreeId/NodeInfo hashing also look like possible candidates to improve. Would be interested in seeing any impovements you manage to make here.

cjpearson:
@ajcvickers Thanks for the feedback.

Making `MaxNodeCountForTransformations` configurable sounds like a good change to me. It would solve our issues and seems like the least risky option.

The BitVec implementation is almost exactly the same as the .NET implementation, so I'm pretty confident of its behavior. The only major change besides exposing the bits was trying to use an ArrayPool, but I'm not sure the benefits were worth the risk there. Subsumes and MoveNext were the only parts of VarVec that changed so more tests there might help. 

Unfortunately the hashing turned into a dead end for me. Any improvements I found also broke the compilation. I don't think there's any quick improvements here. 

Do you have any thoughts on the VarMap changes?

chrfin:
I have/had the same problem. I merged your changes into the 6.2.0 branch and my query compile time went down from ~4 minutes to ~10s :-)

chrfin:
@ajcvickers Will any of this be available in EF 6.3?

ajcvickers:
@chrfin I plan to see what I can merge, so I hope so, but that's not a guarantee.

chrfin:
@ajcvickers Thanks for your reply. At least the configuration option for `MaxNodeCountForTransformations ` would already help me A LOT. Is this planned from your side or would it make sense to create a pull-request with this property configurable?

chrfin:
@ajcvickers Any update on this with EF 6.3 Preview 6?

ajcvickers:
@chrfin No update for preview 6.

ajcvickers:
@cjpearson Thanks for the PR and sorry for the delay in merging. This is now in.

chrfin:
@ajcvickers  Is this available in a (signed) nightly somewhere? https://www.myget.org/gallery/aspnetwebstacknightly is 6 month old...

ajcvickers:
@bricelam Nightly builds for EF6?

chrfin:
@ajcvickers Yes - including this merge and a valid signature...

bricelam:
https://dotnetfeed.blob.core.windows.net/aspnet-entityframework6/index.json
(may also need https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json)

drauch:
Is this released yet?

Suchiman:
> Is this released yet?

This has been released as part of the 6.4.4 release in 2020-05-12.

PR #440, opened 2018-01-08
Removed legacy support for HistoryRepository

rzakirov:
Removed legacy support for HistoryRepository (it throws exception when uses DbMigrator.Update)


rzakirov:
Sorry, this fix for me only. Pull request is created by mistake

PR #466, opened 2018-02-14
Enforce null semantics for >, >=, <, and <= in plan compiler.

maumar:
ported from CodePlex


ajcvickers:
@divega Talked to @maumar about this and we agreed that these changes are not correct, specifically in negation. Given this is likely a breaking change even if done right I am going to close this for now. We can always revisit if there is interest.

PR #467, opened 2018-02-14
Improve TruncateTime canonical function performance

maumar:
Ported from CodePlex

### DIFF
--- a/src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
+++ b/src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
@@ -1285,49 +1285,63 @@ namespace System.Data.Entity.SqlServer.SqlGen
 
         // <summary>
         // TruncateTime(DateTime X)
-        // PreKatmai:    TRUNCATETIME(X) => CONVERT(DATETIME, CONVERT(VARCHAR(255), expression, 102),  102)
-        // Katmai:    TRUNCATETIME(X) => CONVERT(DATETIME2, CONVERT(VARCHAR(255), expression, 102),  102)
+        // PreKatmai:    TRUNCATETIME(X) => DATEADD(d, DATEDIFF(d, 0, expression), 0)
+        // Katmai:       TRUNCATETIME(X) => CAST(CAST(expression AS DATE) as DATETIME2)
         // TruncateTime(DateTimeOffset X)
-        // TRUNCATETIME(X) => CONVERT(datetimeoffset, CONVERT(VARCHAR(255), expression,  102)
-        // + ' 00:00:00 ' +  Right(convert(varchar(255), @arg, 121), 6),  102)
+        // Katmai only: TRUNCATETIME(X) => TODATETIMEOFFSET(CAST(expression AS DATE), DATEPART(tz, expression)))
         // </summary>
         private static ISqlFragment HandleCanonicalFunctionTruncateTime(SqlGenerator sqlgen, DbFunctionExpression e)
         {
             //The type that we need to return is based on the argument type.
-            string typeName = null;
-            var isDateTimeOffset = false;
-
             var typeKind = e.Arguments[0].ResultType.GetPrimitiveTypeKind();
+            var isDateTimeOffset = (typeKind == PrimitiveTypeKind.DateTimeOffset);
 
-            if (typeKind == PrimitiveTypeKind.DateTime)
-            {
-                typeName = sqlgen.IsPreKatmai ? "datetime" : "datetime2";
-            }
-            else if (typeKind == PrimitiveTypeKind.DateTimeOffset)
-            {
-                typeName = "datetimeoffset";
-                isDateTimeOffset = true;
-            }
-            else
+            if (!isDateTimeOffset && typeKind != PrimitiveTypeKind.DateTime)
             {
                 Debug.Assert(true, "Unexpected type to TruncateTime" + typeKind.ToString());
             }
 
-            var result = new SqlBuilder();
-            result.Append("convert (");
-            result.Append(typeName);
-            result.Append(", convert(varchar(255), ");
-            result.Append(e.Arguments[0].Accept(sqlgen));
-            result.Append(", 102) ");
 
-            if (isDateTimeOffset)
+            var result = new SqlBuilder();
+
+            if (sqlgen.IsPreKatmai)
             {
-                result.Append("+ ' 00:00:00 ' +  Right(convert(varchar(255), ");
+                if (isDateTimeOffset)
+                {
+                    throw new NotSupportedException(Strings.SqlGen_CanonicalFunctionNotSupportedPriorSql10(e.Function.Name));
+                }
+
+                result.Append("dateadd(d, datediff(d, 0, ");
                 result.Append(e.Arguments[0].Accept(sqlgen));
-                result.Append(", 121), 6)  ");
+                result.Append("), 0)");
+            }
+            else
+            {
+                if (!isDateTimeOffset)
+                {
+                    result.Append("cast(");
+                }
+                else
+                {
+                    result.Append("todatetimeoffset(");
+                }
+
+                result.Append("cast(");
+                result.Append(e.Arguments[0].Accept(sqlgen));
+                result.Append(" as date)");
+
+                if (!isDateTimeOffset)
+                {
+                    result.Append(" as datetime2)");
+                }
+                else
+                {
+                    result.Append(", datepart(tz, ");
+                    result.Append(e.Arguments[0].Accept(sqlgen));
+                    result.Append("))");
+                }
             }
 
-            result.Append(",  102)");
             return result;
         }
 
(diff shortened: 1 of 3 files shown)


src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
```diff
@@ -1285,49 +1285,63 @@ private static void AppendConvertToVarchar(SqlGenerator sqlgen, SqlBuilder resul
...
-            result.Append(", convert(varchar(255), ");
-            result.Append(e.Arguments[0].Accept(sqlgen));
-            result.Append(", 102) ");
 
-            if (isDateTimeOffset)
+            if (sqlgen.IsPreKatmai)
+            {
+                if (isDateTimeOffset)
+                {
+                    throw new NotSupportedException(Strings.SqlGen_CanonicalFunctionNotSupportedPriorSql10(e.Function.Name));
```
ajcvickers: This looks like a break.

divega:
I believe the original CodePlex PR was submitted by @brandondahler, in case he is interested in following up.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
brandondahler: For PreKatamai, the sql produced is `CONVERT(datetimeoffset, CONVERT(VARCHAR(255), expression,  102) + ' 00:00:00 ' +  Right(convert(varchar(255), @arg, 121), 6),  102)`, which is invalid for PreKatamai because the `datetimeoffset` data type didn't exist in Sql Server < 2008.

Because of this I don't believe that this is actually a break.

brandondahler:
Thanks for the tag, answered inline.

ajcvickers:
@divega This look okay to you?

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
```diff
@@ -1285,49 +1285,63 @@ private static void AppendConvertToVarchar(SqlGenerator sqlgen, SqlBuilder resul
 
         // <summary>
         // TruncateTime(DateTime X)
-        // PreKatmai:    TRUNCATETIME(X) => CONVERT(DATETIME, CONVERT(VARCHAR(255), expression, 102),  102)
-        // Katmai:    TRUNCATETIME(X) => CONVERT(DATETIME2, CONVERT(VARCHAR(255), expression, 102),  102)
+        // PreKatmai:    TRUNCATETIME(X) => DATEADD(d, DATEDIFF(d, 0, expression), 0)
+        // Katmai:       TRUNCATETIME(X) => CAST(CAST(expression AS DATE) as DATETIME2)
```
divega: There appears to be no significant perf advantage advantage in the Katmai translation over the PreKatmai translaiton. I tested it with:

``` sql
declare @expression datetime;
declare @ret datetime2;

set @expression = CAST('1969/08/22 10:10' as DATETIME2);


DECLARE @Iteration INT
SET @Iteration = 1
WHILE @Iteration <= 100000000
BEGIN

	--SET @ret = CONVERT(DATETIME, convert(varchar(255), @expression, 102), 102); -- 92 seconds
	--SET @ret = DATEADD(d, DATEDIFF(d, 0, @expression), 0); -- 55 seconds
	SET @ret = CAST(CAST(@expression AS DATE) as DATETIME2); -- 54 seconds

    SET @Iteration = @Iteration + 1

END

print @ret
print @iteration
```
I would suggest we simplify.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
```diff
@@ -1285,49 +1285,63 @@ private static void AppendConvertToVarchar(SqlGenerator sqlgen, SqlBuilder resul
...
+                }
+
+                result.Append("dateadd(d, datediff(d, 0, ");
+                result.Append(e.Arguments[0].Accept(sqlgen));
+                result.Append("), 0)");
+            }
+            else
             {
-                result.Append("+ ' 00:00:00 ' +  Right(convert(varchar(255), ");
+                if (!isDateTimeOffset)
```
divega: Prefer code clarity: I don't think the few lines that are common are worth the entanglement of the two translations.

test/EntityFramework/FunctionalTests/Query/SqlCeCanonicalFunctionsTests.cs
```diff
@@ -514,7 +514,7 @@ public void CanonicalFunction_truncatetime_translated_properly_to_sql_function()
             using (var context = GetArubaCeContext())
             {
                 var query = context.CreateQuery<DateTime>(@"SELECT VALUE Edm.TruncateTime(A.c5_datetime) FROM ArubaCeContext.AllTypes AS A");
-                Assert.Contains("CONVERT", query.ToTraceString().ToUpperInvariant());
+                Assert.Contains("DATEADD(D, DATEDIFF(D,", query.ToTraceString().ToUpperInvariant());
```
divega: We only have canonical function tests for SQL CE? :man_facepalming: Glad we have coverage for this in https://github.com/aspnet/EntityFramework6/blob/master/test/EntityFramework/FunctionalTests/ProductivityApi/DbFunctionScenarios.cs.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
divega: I agree this should be fine. It is also possible that this just caused an error on the server, but it is also possible that it is blocked by something upstream in the query pipeline.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
brandondahler: Tested this myself and got very similar results with your test, but found much less CPU time was used on the Katmai translation based on the following script:

```
-- Setup 
CREATE TABLE #test (Value datetime2)

SET NOCOUNT ON;

DECLARE @Iteration INT
SET @Iteration = 1
WHILE @Iteration <= 10000000
BEGIN
    INSERT INTO #Test
    SELECT DATEADD(mi, @Iteration, CAST('1969/08/22 10:10' as DATETIME2))


    SET @Iteration = @Iteration + 1
    IF CAST((@Iteration / 100000) as int) * 100000 = @Iteration
       print @iteration

END


-- Test
SET STATISTICS TIME ON;
SET STATISTICS IO ON;

-- Table '#Test__...'. Scan count 9, logical reads 22076, physical reads 0, read-ahead reads 0, lob logical reads 0, lob physical reads 0, lob read-ahead reads 0.
--
-- SQL Server Execution Times:
--   CPU time = 1423 ms,  elapsed time = 221 ms.
SELECT 1 from #test where CAST(CAST(value AS DATE) as DATETIME2) = CAST('2019-01-05' as datetime2)

--Table '#Test__...'. Scan count 9, logical reads 22076, physical reads 0, read-ahead reads 0, lob logical reads 0, lob physical reads 0, lob read-ahead reads 0.
--
-- SQL Server Execution Times:
--   CPU time = 2844 ms,  elapsed time = 379 ms.
SELECT 1 from #test where DATEADD(d, DATEDIFF(d, 0, Value), 0) = CAST('2019-01-05' as datetime2)


SET STATISTICS TIME OFF;
SET STATISTICS IO OFF;


-- Cleanup
DROP TABLE #test;
```

Based on that test it appears that it might be as much as 50% less cpu-intensive to use the Katmai version over the Pre-Katmai version.  I don't particular have an explanation for why the results are different from your test's results elapsed-time wise.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
```diff
@@ -1285,49 +1285,63 @@ private static void AppendConvertToVarchar(SqlGenerator sqlgen, SqlBuilder resul
...
-            }
-            else if (typeKind == PrimitiveTypeKind.DateTimeOffset)
-            {
-                typeName = "datetimeoffset";
-                isDateTimeOffset = true;
-            }
-            else
+            if (!isDateTimeOffset && typeKind != PrimitiveTypeKind.DateTime)
             {
                 Debug.Assert(true, "Unexpected type to TruncateTime" + typeKind.ToString());
```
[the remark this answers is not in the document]
brandondahler: Funnily enough, `Debug.Assert(true, ...) ` is also a GNDN.  Will be properly fixed as well.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
divega: Very cool! I can confirm your findings.

src/EntityFramework.SqlServer/SqlGen/SqlFunctionCallHandler.cs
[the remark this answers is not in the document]
divega: Missed that.

brandondahler:
Changes made in https://github.com/brandondahler/EntityFramework6/tree/truncate_time, unable to actually update this pull request because it migrated from @maumar from CodePlex.

Let me know if you'd like me to open a new PR or if one of the members of the repository will take care of it.  Was able to find a good place to put a net-new test for the updated code.

divega:
I think a new PR would be good. That way the commits will be correctly assigned to you.

divega:
Superseded by https://github.com/aspnet/EntityFramework6/pull/948.

PR #481, opened 2018-02-28
Accessibility Bugs 457911 and 457912

lajones:
Fixes for Accessibility bugs 457911 & 457912.

Also an extra fix for 458689 - which now will have the errors Narrated for the Reference External Type field (in addition to the existing fix for the Enum Type Name field).


lajones:
Merged into master with 27065f255.

PR #482, opened 2018-03-01
Fix for Accessibility bug 458753. Help text for Mapping Details Window.

lajones:
Adding help text for Mapping Details Window.


lajones:
Merged into master with commit 3ff2c134a.

PR #483, opened 2018-03-03
Fix for Accessibility Bug 457906 - visible warning on Add New Enum dialog

lajones:
When either the "Enum Type Name" or "Reference External Type" textboxes have errors, the error now shows up with warning image and popup containing the text of the error.


lajones:
Merged into master with commit 05285da32.

PR #484, opened 2018-03-03
Follow-up Fix for Accessibility Bug 457906

lajones:
See https://github.com/aspnet/EntityFramework6/pull/483 - forgot to update the XAML files for the other versions of VS.


lajones:
Merged into master with commit 73b3e3d90.

Issue #493, opened 2018-03-15
Dml statements containing DbNotExpression cause StackOverflowException


sshukurov:
In my application I have a db interceptor where I am passing a DbUpdateCommandTree with DbNotExpression in predicate. Execution of such command tree leads to StackOverflowException.

I investigated the problem cause, obviously there is a bug in DmlSqlGenerator.cs:
In method ExpressionTranslator.Visit(DbNotExpression expression) instead of:
`expression.Accept(this);`
there should be:
`expression.Argument.Accept(this);`

PR #494, opened 2018-03-15
Fix for bug: DbUpdateCommandTree with DbNotExpression causing StackOverflowException

sshukurov:
Issue #493


src/EntityFramework.SqlServer/SqlGen/DmlSqlGenerator.cs
```diff
@@ -807,6 +808,37 @@ var variableAssignment
                 }
             }
 
+            public override void Visit(DbInExpression e)
```
sshukurov: Code was adapted from SqlGenerator.cs.

PR #503, opened 2018-03-20
Fix for issue 4 - improve perf of metadata queries on SQL Server

lajones:
Added an interceptor which appends "OPTION (MERGE JOIN)" to the SQL for the metadata queries executed during Reverse Engineering. Added a quirk so users can turn this off if they want to. (I thought this would be best, but let me know if you'd prefer the quirk to work the other way around i.e. users having to explicitly set an `AppSetting` to turn this feature on).

### DIFF
--- a/src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
+++ b/src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
@@ -24,7 +24,10 @@ namespace Microsoft.Data.Entity.Design.VersioningFacade.ReverseEngineerDb.Schema
         private readonly IDbCommandInterceptor _addOptionMergeJoinInterceptor;
 
         [SuppressMessage("Microsoft.Usage", "CA1806:DoNotIgnoreMethodResults", MessageId = "System.Boolean.TryParse(System.String,System.Boolean@)")]
-        public EntityStoreSchemaGeneratorDatabaseSchemaLoader(EntityConnection entityConnection, Version storeSchemaModelVersion)
+        public EntityStoreSchemaGeneratorDatabaseSchemaLoader(
+            EntityConnection entityConnection,
+            Version storeSchemaModelVersion,
+            bool needsMergeJoinHint)
         {
             Debug.Assert(entityConnection != null, "entityConnection != null");
             Debug.Assert(entityConnection.State == ConnectionState.Closed, "expected closed connection");
@@ -46,11 +49,9 @@ namespace Microsoft.Data.Entity.Design.VersioningFacade.ReverseEngineerDb.Schema
                         metadataMergeJoinsAppSetting, out switchOffMetadataMergeJoins);
                 }
             }
-            if (!switchOffMetadataMergeJoins
-                && _connection != null
-                && typeof(System.Data.SqlClient.SqlConnection) == _connection.StoreConnection.GetType())
+            if (!switchOffMetadataMergeJoins && needsMergeJoinHint)
             {
-                    _addOptionMergeJoinInterceptor = new AddOptionMergeJoinInterceptor();
+                _addOptionMergeJoinInterceptor = new AddOptionMergeJoinInterceptor();
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
```diff
@@ -26,6 +32,26 @@ public EntityStoreSchemaGeneratorDatabaseSchemaLoader(EntityConnection entityCon
...
+                    metadataMergeJoinsAppSettings.FirstOrDefault();
+                if (metadataMergeJoinsAppSetting != null)
+                {
+                    bool.TryParse(
+                        metadataMergeJoinsAppSetting, out switchOffMetadataMergeJoins);
+                }
+            }
+            if (!switchOffMetadataMergeJoins
+                && _connection != null
+                && typeof(System.Data.SqlClient.SqlConnection) == _connection.StoreConnection.GetType())
```
ajcvickers: Not sure that this is the best way to test for SQL Server because when a wrapping provider is being used the connection type can be something different than SqlConnection. Is there something in the SSDL metadata that indicates SQL Server?

src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
```diff
@@ -487,5 +527,46 @@ ORDER BY
...
             ,   sp.Ordinal
             ";
+
+        /// <summary>
+        /// This interceptor is added to work around a problem with SQL Server
+        /// cardinality estimation: https://github.com/aspnet/EntityFramework6/issues/4
+        /// which in turn causes bad performance running the queries above on large models.
+        /// We can workaround this by directing SQL Server to use MERGE JOINs.
+        /// </summary>
+        private sealed class AddOptionMergeJoinInterceptor : IDbCommandInterceptor
```
ajcvickers: I think there is probably an abstract base class that could be used here instead of using the interface directly. Can you check?

lajones: There is. Fixed in next commit.

src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
lajones: Understood. Two approaches I can take here:
1. I can look for the string "provider=System.Data.SqlClient" in the existing `entityConnection.ConnectionString`.
2. I can pass in the `ModelBuilderSettings` object (which has all the information we've gathered about the provider, including e.g. the `RuntimeProviderInvariantName`) into the constructor of `EntityStoreSchemaGeneratorDatabaseSchemaLoader`.

Option 1 will probably need a regex - which we were trying to avoid - but we'd need to make sure we'd reached the end of the string or that the next character is a separator character.
Option 2 implies changing a public constructor but of an internal class - so not a breaking change.

I think I prefer option 2. Are you OK with that?

lajones: Actually the 2nd approach causes `EntityDesignerVersioningFacade` to have a dependency on `EntityDesign` which it should not have. However there is a 3rd approach -  can just pass a `bool` saying whether we need to provide the MERGE JOIN hint. That way the caller (who has the information) can decide. I'll include that in the next commit.

src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
```diff
@@ -15,17 +18,44 @@ namespace Microsoft.Data.Entity.Design.VersioningFacade.ReverseEngineerDb.Schema
...
         {
             Debug.Assert(entityConnection != null, "entityConnection != null");
             Debug.Assert(entityConnection.State == ConnectionState.Closed, "expected closed connection");
             Debug.Assert(EntityFrameworkVersion.IsValidVersion(storeSchemaModelVersion), "invalid version");
 
             _connection = entityConnection;
             _storeSchemaModelVersion = storeSchemaModelVersion;
+
+            if (needsMergeJoinHint)
+            {
```
divega: can we extract all the quirk logic into its own private method for clarity?

src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
```diff
@@ -15,17 +18,44 @@ namespace Microsoft.Data.Entity.Design.VersioningFacade.ReverseEngineerDb.Schema
...
         private readonly EntityConnection _connection;
         private readonly Version _storeSchemaModelVersion;
+        private readonly IDbCommandInterceptor _addOptionMergeJoinInterceptor;
 
-        public EntityStoreSchemaGeneratorDatabaseSchemaLoader(EntityConnection entityConnection, Version storeSchemaModelVersion)
+        [SuppressMessage("Microsoft.Usage", "CA1806:DoNotIgnoreMethodResults", MessageId = "System.Boolean.TryParse(System.String,System.Boolean@)")]
+        public EntityStoreSchemaGeneratorDatabaseSchemaLoader(
+            EntityConnection entityConnection,
+            Version storeSchemaModelVersion,
+            bool needsMergeJoinHint)
```
divega: I don't think we need to lift the decision to the ModelGenerator and add this parameter. Another alternative to the ones you mentioned before is to obtain the invariant name from `_connection.StoreProviderFactory.GetProviderInvariantName()`, so you actually already have the information you need in this class.

ajcvickers:
Looks good with the changes that @divega suggested. I'll leave Diego to do the final sign-off.

src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
lajones: Done.

src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs
lajones: Fixed. That's much better. (Btw it's actually `_connection.StoreConnection.GetProviderInvariantName()`).

lajones:
All changed wanted by @divega are now fixed. Going ahead with the merge.

lajones:
Merged into master with fb2695c2b.

PR #506, opened 2018-03-27
Fixes for Accessibility Bugs 458661 & 586335

lajones:
Fixes for Accessibility bugs 458661 and 586335. 

The fix for 458661 changes the Add/Edit Enum Type Dialog so that the help text is available on every cell of the data grid as requested.

I am unable to fix 586335 the way they suggest in the bug - the underlying layer does not expose the required API. However I added handlers for Ctrl+Up and Ctrl+Down which cause the appropriate entity or entity compartment to collapse or expand. Then I added help text that should make this clear to a user. So that should provide a solution for users who use screen readers.

### DIFF
--- a/src/EFTools/EntityDesign/Resources.Designer.cs
+++ b/src/EFTools/EntityDesign/Resources.Designer.cs
@@ -820,6 +820,15 @@ namespace Microsoft.Data.Entity.Design {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to Start typing or press &apos;F2&apos; to edit.
+        /// </summary>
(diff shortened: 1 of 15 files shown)


ajcvickers:
@lajones When you say, "underlying layer" do you mean the DSL Toolkit?

lajones:
@ajcvickers yes.

ajcvickers:
@lajones If there is further pushback on 586335, then we can route to the DSL Tookit team.
/cc @divega

lajones:
Checked into master with 72b5214fb.

PR #511, opened 2018-03-29
Fix for Accessibility Bug 586335

lajones:
Fix for Accessibility Bug 586335 - providing key handlers for EntityType shape and Properties & Navigation Properties compartments.

### DIFF
--- a/src/EFTools/EntityDesignEntityDesigner/CustomCode/Shapes/EntityTypeShape.cs
+++ b/src/EFTools/EntityDesignEntityDesigner/CustomCode/Shapes/EntityTypeShape.cs
@@ -33,7 +33,6 @@ namespace Microsoft.Data.Entity.Design.EntityDesigner.View
         private static Image _complexPropertyBitmap;
         private static Image _navigationPropertyBitmap;
         private bool _isAdjustedForFillColor;
-        private EntityChevronButtonField _expandCollapseField;
 
         internal class EntityChevronButtonField : ChevronButtonField
         {
@@ -104,6 +103,9 @@ namespace Microsoft.Data.Entity.Design.EntityDesigner.View
             bounds.Location = randomized;
             AbsoluteBounds = bounds;
             EntityDesignerViewModel.EntityShapeLocationSeed++;
+
+            // Provide keyboard shortcuts for collapse and expand of this EntityType shape
+            this.KeyUp += EntityTypeShape_KeyUp;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/EFTools/EntityDesignEntityDesigner/CustomCode/Shapes/EntityTypeShape.cs
```diff
@@ -690,28 +687,35 @@ public override string AccessibleHelp
...
         }
 
-        public override void OnKeyUp(DiagramKeyEventArgs e)
+        private void EntityTypeShape_KeyUp(object sender, DiagramKeyEventArgs e)
         {
             if (e.Control && e.KeyCode == Keys.Up)
             {
-                var isExpanded = _expandCollapseField.IsExpanded(this);
-                if (isExpanded.HasValue && isExpanded.Value)
+                if (IsExpanded)
```
divega: Nit: I feel tempted to try to refactor this to have a single extension method, e.g. something like EnsureExpandedState(this NodeShape shape, bool state) that could be reused for compartments and entities. Not sure it is possible, and it is not really necessary, just a matter of reducing code duplication.

lajones: Agreed. I'll try to find a good place to put it and call it. But if it ends up taking too long I'll just revert to this. I don't want to spend too much time on it, this way is clear to read and it's unlikely we will have to revisit or generalize the collapse/expand code any time in the future.

src/EFTools/EntityDesignEntityDesigner/CustomCode/Utils/NodeShapeExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    using (var txn = nodeShape.Store.TransactionManager.BeginTransaction())
+                    {
+                        nodeShape.IsExpanded = false;
+                        txn.Commit();
+                    }
+                }
+            }
+        }
+    }
```
divega: @lajones 
I think you should be able to reduce the whole method to something like:
``` C#
public static void EnsureExpandedState(this NodeShape nodeShape, bool expanded)
{
    if (nodeShape.IsExpanded != expanded)
    {
        using(var txn = nodeShape.Store.TransactionManager.BeginTransaction())
        {
            nodeShape.IsExpanded = expanded;
            txn.Commit();
        }
    }
}
```

lajones:
Checked into master with 282ef81a2

src/EFTools/EntityDesignEntityDesigner/CustomCode/Utils/NodeShapeExtensions.cs
lajones: Yes, that does look better, but I've already submitted the build and it doesn't make a huge difference - so leaving as it is.

lajones: Note: updated in #517

PR #517, opened 2018-04-06
Fix for Accessibility bug 586335. 

lajones:
Update help text to describe the collapsed/expanded current state on the EntityTypeShape and its compartments, and add rename instructions for the EntityTypeShape.


lajones:
Merged into master with 22a11d132

Issue #535, opened 2018-05-11
EF 6.2 tooling in VS 15.7 breaks SQLite EDMX generation


ErikEJ:
2 of my users have seen issues with 15.7 - am I correct that the 15.7 release finally contains the EF 6.2 tooling?

See detailed repro steps here: https://github.com/ErikEJ/SqlCeToolbox/issues/710

### Further technical details
EF version: EF 6
Database Provider: SQLite
IDE: Visual Studio 2017 15.7

divega:
Yes @ErickEJ, that is correct.

divega:
@lajones, @ajcvickers it seems that the call to [GetProviderInvarianName](https://github.com/aspnet/EntityFramework6/blame/master/src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs#L44) we introduced as one of the improvements throws for SQL CE, at least in some cases. Perhaps this is due to the provider being configured for the application but not for VS?

ErikEJ:
It is for SQLite, have not tested with SQL CE, but I will

divega:
Thanks for clarifying. My brain’s lookup tables contain some stale entries :smile:

Perhaps the connection.GetProviderInvarianName() extension method operates on EF configuration. I think we have another version of the method on DbPrividerFactoy. Not sure if it has the same issue.

ErikEJ:
Just tested with SQLCE, and it is working

ajcvickers:
Assigning this to @lajones. I think we can be more defensive in calling `connection.GetProviderInvarianName()` and if it throws, then assume that it's not SQL Server. We may want to do a 6.2.1 patch for this.

ErikEJ:
Any suggestions for workarounds?

lajones:
Agreed on the defensive calling. But for now it looks as though it's having problems with reading the something in the `<system.data><DbProviderFactories>` section of the `app.config`. Or maybe it's setting up it's cache incorrectly. Are the `<remove>` entries necessary? Can they be removed, or moved so that all `<remove>` entries are before all the `<add>` ones? I wonder if somehow removing `System.Data.SQLite` is affecting `System.Data.SQLite.EF6` as well?

ajcvickers:
[...]

PR #540, opened 2018-05-17
Fix for issue 535. Ignore exceptions from GetProviderInvariantName()

lajones:
In [EntityStoreSchemaGeneratorDatabaseSchemaLoader](https://github.com/aspnet/EntityFramework6/blob/22a11d132306c257435fa7d6d75a31359cbbd05a/src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs#L21) we attempt to determine if the provider is SqlServer. In certain cases even attempting to determine this can throw various kinds of exception. If it throws then not catching the exception results in the tool dying and the exception reported in the Output Window. Since all we want to know here is whether the provider is SqlServer this update just assumes that the provider is not SqlServer if there is an exception.


lajones:
Fix for #535 checked into tooling-release-15.8 with commit 091b7a2. Note: will need to wait for future VS check-in for this to appear in VS.

PR #546, opened 2018-05-23
ObjectMaterializedEventArgs constructor maked public.

svishnevsky:
The reason of this change is unit testing ObjectMaterialized event handler. While ObjectMaterializedEventArgs has a internal constructor, it's impossible to instantiate this class into Unit Test that was written for a ObjectMaterialized event handler of the ObjectContext.


ajcvickers:
Merged. Thanks!

Issue #530, opened 2018-05-06
Poor handling of DbNull in FieldDescriptor


denwanke:
Consider the following (artificial) code snippet:
```
var query = "SELECT Customer.Name, ANYELEMENT(Customer.Orders) FROM CustomerData.Customers AS Customer";
var data = context.CreateQuery<object>(query).Execute(MergeOption.NoTracking);
```
When `data` gets data-bound in a data-consuming component (from a third-party lib) and there is no orders for a particular customer, the following code line is triggered in `FieldDescriptor.GetValue()`:

`throw new ArgumentException(Strings.ObjectView_IncompatibleArgument);`

The is caused by the check `_itemType.IsAssignableFrom(item.GetType())` being negative when `item` is a `DbNull`. Which in turn is caused by `GetValue()` returning a `DbNull` for the second field of the data record and the data-bound component getting confused with it, mistakenly considering it a non-empty value.

The possible fix could be as easy as placing a test for `DbNull` right between the two precondition checks in `GetValue()`: 
```

public override object GetValue(object item)
{
    Check.NotNull(item, "item");

    if (item is DBNull) return item; <-- protection against DbNull

    if (!_itemType.IsAssignableFrom(item.GetType()))
    {
        throw new ArgumentException(Strings.ObjectView_IncompatibleArgument);
    }
    ...
}
```

Even better would be to have `GetValue()` to always return `null` instead of `DbNull`:

```
public override object GetValue(object item)
{
    ...
    return propertyValue is DBNull ? null : propertyValue;
}
```

Would it be a breaking change?

ajcvickers:
@denwanke Yes, both suggestions would be a breaking changes.

ajcvickers:
**EF Team Triage:** This issue is not something that our team is planning to address in the EF6.x code base.  This does not mean that we would not consider a community contribution to address this issue.

Moving forwards, our team will be fixing bugs, implementing small improvements, and accepting community contributions to the EF6.x code base. Larger feature work and innovation will happen in the EF Core code base (https://github.com/aspnet/EntityFramework).

**Closing an issue in the EF6.x project does not exclude us addressing it in EF Core. In fact, a number of popular feature requests for EF have already been implemented in EF Core (alternate keys, batching in SaveChanges, etc.).**

[...]

PR #551, opened 2018-06-10
Issue #530: improved handling of DbNull in FieldDescriptor

denwanke:
Fix proposal for issue #530


ajcvickers:
@denwanke Can you explain in some detail why this change is needed? Also, please add tests that demonstrate what was broken before and verify that it is fixed after this change.

ajcvickers:
Closing due to inactivity.

Issue #531, opened 2018-05-06
Missing support for EDM collections in DataRecordObjectView


denwanke:
Consider the following (artificial) code snippet:

```
var query = "SELECT customer.Name, (SELECT order.Date, order.Summary FROM customer.Orders as order) AS Orders FROM CustomerData.Customers AS customer";
var data = context.CreateQuery<object>(query).Execute(MergeOption.NoTracking);
```
When data gets data-bound in a hierarchical data-consuming component (from a third-party lib), the method `ITypedList.GetItemProperties()` returns no property descriptors for the nested (order) rows of a particular customer. 

In this case the parameter `listAccessors` contains the FieldDescriptor for the second field of the main data record but its real EDM type is nested in an EDM collection.

The following fix is suggested:
```
PropertyDescriptorCollection ITypedList.GetItemProperties(PropertyDescriptor[] listAccessors)
{
    if (listAccessors == null || listAccessors.Length == 0)
    {
        // Caller is requesting property descriptors for the root element type.
        return _propertyDescriptorsCache;
    }

    Type itemType = null;

    // Use the last PropertyDescriptor in the array to build the collection of returned property descriptors.
    var propertyDescriptor = listAccessors[listAccessors.Length - 1];
    var fieldDescriptor = propertyDescriptor as FieldDescriptor;
    if (fieldDescriptor != null && fieldDescriptor.EdmProperty != null)
    {
        var edmType = fieldDescriptor.EdmProperty.TypeUsage.EdmType;

        // Unwrap collection types
        if (edmType.BuiltInTypeKind == BuiltInTypeKind.CollectionType)
        {
            edmType = ((CollectionType)edmType).TypeUsage.EdmType;
        }

        // If the property descriptor describes a data record with the EDM type of RowType,
        // construct the collection of property descriptors from the property's EDM metadata.
        if (edmType.BuiltInTypeKind == BuiltInTypeKind.RowType)
        {
            // Retrieve property descriptors from EDM metadata.
            return MaterializedDataRecord.CreatePropertyDescriptorCollection((RowType)edmType, typeof(IDataRecord), true);
        }

        itemType = edmType.ClrType;
    }
            
    // Use the CLR type.

    if (itemType == null)
    {
        itemType = GetListItemType(propertyDescriptor.PropertyType);
    }

    return TypeDescriptor.GetProperties(itemType);
}
```
[...]

PR #552, opened 2018-06-10
Issue #531: support for EDM collections in DataRecordObjectView

denwanke:
Fix proposal for issue #531


ajcvickers:
@denwanke Can you please describe the scenario by which an application can hit this issue, and also add tests that demonstrate this and verify that the issue is no longer present after the fix?

ajcvickers:
Closing due to inactivity.

PR #590, opened 2018-08-08
Allow later versions of Microsoft.SqlServer.Types to be loaded

ajcvickers:
Still looks for 11 then 10 first to avoid breaking, and if not found attempts to load the most recent version from 20 down.

### DIFF
--- a/src/EntityFramework.SqlServer/SqlTypesAssemblyLoader.cs
+++ b/src/EntityFramework.SqlServer/SqlTypesAssemblyLoader.cs
@@ -5,12 +5,16 @@ namespace System.Data.Entity.SqlServer
     using System.Collections.Generic;
     using System.Data.Entity.SqlServer.Resources;
     using System.Diagnostics.CodeAnalysis;
+    using System.Globalization;
     using System.IO;
     using System.Linq;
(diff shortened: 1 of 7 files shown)


ajcvickers:
@bricelam Sorry to randomize with EF6 stuff!

bricelam:
When I dug into this a few months ago, I came to the conclusion that SqlClient will only work if you redirect newer versions back down 11.0.0.0...

bricelam:
Here's [the code](https://referencesource.microsoft.com/#System.Data/System/Data/SqlClient/SqlConnectionString.cs,412) that assumes either 10.0.0.0 or 11.0.0.0.

ajcvickers:
@bricelam (Most) tests are passing on my machine, which after installing the latest stuff has 14 in the GAC and 14 is being loaded.

ajcvickers:
Updated with additional test fixes. Also, updated database existence checks to remove tests that were using db_id() since it no longer sees other database when permissions to view master have been removed. This seems like an intentional change on the part of SQL Server. It doesn't break any mainline cases in EF6, and doesn't change the Azure behavior.

PR #627, opened 2018-09-17
Prevents the previously assigned transaction from being deleted due t…

klauzyx:
I'm currently working with a specific DB access driver, and I have found the problem that the command object, at the moment the connection is assigned, puts null in the transaction it had previously assigned.

As I don't know the reason why the transaction is assigned before the connection, I have proceeded to review the assignment of the transaction after the assignment of the connection.


klauzyx:
Sorry, I just noticed that I made the modification in the Execute method, but not in its respective ExecuteAsync method.

The pull request is already started, how do I proceed?

ajcvickers:
@klauzyx This looks like a breaking change in that it changes the way command.Transaction is assigned. We would only take a change of this kind if there is very high confidence that it won't break other providers. Have you tested against other common providers?

ajcvickers:

Issue #591, opened 2018-08-13
Indexes are not working as intended


pavlexander:
There are at least 3 issues with "Index" functionality which I would like to report. I hope you don't mind having them in 1 place.

### Steps to reproduce (issue 1)
I am trying to create 2 indexes like this:

```c#
modelBuilder.Entity<MyEntity>()
    .HasIndex(p => new { p.Column1, p.Column2, p.Column3, p.Column4 })
    .HasName("ix_index1")
    .IsUnique();

modelBuilder.Entity<MyEntity>()
    .HasIndex(p => new { p.Column1, p.Column2, p.Column3 })
    .HasName("ix_index2")
    .IsUnique();
```

Upon running command add-migration InitialCreate what I get in return, is this script:

```c#
CreateTable(
    "DEV.MyEntity",
    c => new
        {
            Column1 = c.String(nullable: false, maxLength: 10),
            Column2 = c.Decimal(nullable: false, precision: 19, scale: 0),
            Column3 = c.Decimal(nullable: false, precision: 10, scale: 0),
            Column4 = c.Decimal(nullable: false, precision: 10, scale: 0),
        })
    .PrimaryKey(t => new { t.Column1, t.Column2, t.Column3 })
    .Index(t => new { t.Column1, t.Column2, t.Column3 }, unique: true, name: "ix_index2")
    .Index(t => t.Column4, unique: true, name: "ix_index1");
```

Is there a reason why of all columns in index ix_index1 only Column4 is present? The expected result is that ix_index1 is for 4 columns.

### Steps to reproduce (issue 2)

Despite having issue 1 - I was able to create index using another approach:

```c#
modelBuilder.Entity<Invoice>()
	.Property(e => e.ExternalSystemId)
	.HasColumnAnnotation(
		IndexAnnotation.AnnotationName,
		new IndexAnnotation(new[]
		{
			new IndexAttribute("ix_invoice_invoice_id_status", 1) { IsUnique = true }
		}));

modelBuilder.Entity<Invoice>()
	.Property(e => e.ClientId)
	.HasColumnAnnotation(
		IndexAnnotation.AnnotationName,
		new IndexAnnotation(new[]
		{
			new IndexAttribute("ix_invoice_invoice_id_status", 2) { IsUnique = true }
		}));

modelBuilder.Entity<Invoice>()
	.Property(e => e.InvoiceId)
	.HasColumnAnnotation(
		IndexAnnotation.AnnotationName,
[...]

PR #677, opened 2018-12-17
Fixes part of #591; when 2 calls to HasIndex reference same column

phkelley:
I was looking at what's required to move my project from EF 6.2 to EF Core, and one big change is the move from `[Index]` attributes to calls to `.HasIndex`. So I said, let me change to `.HasIndex` first, since that syntax is available in EF 6.2. Then when I switch to EF Core, it'll be a snap - almost no code changes.

But I hit a bug that happens when you use `HasIndex` to establish two indexes that both contain the same column. This bug was previously reported in #591.

Basically, since the data is stored internally in EF by column, a second index definition displaces the first, for those columns that are in both indexes. I have updated the logic to do a merge instead, and added a test case.


Kantuz001:
i hope this will be merged

ajcvickers:
@phkelley Thanks for the PR and sorry for the delay. This is now merged.

Issue #684, opened 2018-12-30
FeatReq: BeforeUp AfterUp BeforeDown AfterDown partial methods in (generated migration name).designer.cs


ericnewton76:
Sometimes need a way to add script or unsupported migration stuff on before or after Up or Down... ie new partial methods BeforeUp AfterUp BeforeDown AfterDown

These partial methods can manually updated and be put into a separate file that won't get overwritten by `Add-migration -force` 

new (migrationname).Designer.cs:
```c#
		public override void Up()
		{
			BeforeUp();
                        MigrateUp(); //add-migration generated code
			AfterUp();
		}

		public override void Down()
		{
			BeforeDown();
                        MigrateDown(); //add-migration generated code
			AfterDown();
		}

		partial void BeforeUp();
		partial void MigrateUp();
		partial void AfterUp();
		partial void MigrateDown();
		partial void BeforeDown();
		partial void AfterDown(); 
```
this will work just fine going forward, since all the other Add-Migration code has been baked in already.  

And then Add-Migration writes out (migrationname).cs
```c#
public void MigrateUp()
{
    //migration statements
}
public void MigrateDown()
{
    //migration statements
}
```

Developer can now add (migrationname).manual.cs
```c#
public partial class (migrationTypeName)
{
	void AfterUp()
	{
		CreateStoredProcedure("sproc_name", "sproc body");
		CreateStoredProcedure("sproc_name2", "sproc body");

		ExecuteSql("UPDATE OldTableWithNewColumn SET NewColumn='some computed value' WHERE 1=1");
                //or if supporting multiple db backends, calls to other types to generate the appropriate sql
	}

	void BeforeDown()
	{
		DropStoredProcedure("sproc_name");
		DropStoredProcedure("sproc_name2");
	}

}
```

### Further technical details
EF version: 6.2.0+
Database Provider: all
Operating system: n/a
IDE: n/a

ajcvickers:
**EF Team Triage:** This issue is not something that our team is planning to address in the EF6.x code base.  This does not mean that we would not consider a community contribution to address this issue.

Moving forwards, our team will be fixing bugs, implementing small improvements, and accepting community contributions to the EF6.x code base. Larger feature work and innovation will happen in the EF Core code base (https://github.com/aspnet/EntityFramework).

[...]

PR #685, opened 2018-12-30
Add partial methods to codegen for migrations

ericnewton76:
After working with and evolving migrations as I develop, I found that the current system is too eagerly walking over potentially hand coded migration code.  Developers are encouraged to use `Add-Migration` and then told again to use `Add-Migration -force` to update a migration.

This helps to deal with this issue, by creating partial methods that can be implemented in a separate file that is not touched by the Add-Migration process, even using -force

- [x] CSharp code gen
- [x] VB code gen

A practical implementation to Issue #684

### DIFF
--- a/src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
+++ b/src/EntityFramework/Migrations/Design/CSharpMigrationCodeGenerator.cs
@@ -98,7 +98,7 @@ namespace System.Data.Entity.Migrations.Design
                         @namespace, className, writer, "DbMigration", designer: false,
                         namespaces: GetNamespaces(operations));
 
-                    writer.WriteLine("public override void Up()");
+                    writer.WriteLine("partial void MigrateUp()");
                     writer.WriteLine("{");
                     writer.Indent++;
 
@@ -112,7 +112,7 @@ namespace System.Data.Entity.Migrations.Design
 
                     writer.WriteLine();
 
-                    writer.WriteLine("public override void Down()");
+                    writer.WriteLine("partial void MigrateDown()");
                     writer.WriteLine("{");
                     writer.Indent++;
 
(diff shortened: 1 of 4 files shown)


ericnewton76:
Wow these code generation tests are miserable.   Need to be using Roslyn CSharpSyntaxTree.Parse()

ericnewton76:
Most of the tests worked on my machine, especially in the areas I touched.  I did get some wierd multiple use errors when running them though.

ajcvickers:
@bricelam Is this something we want to take? (I know we discussed it on Core, but I don't remember the conclusion.)

bricelam:
I started taking this a bit further in [bricelam:quote](https://github.com/aspnet/EntityFramework6/compare/master...bricelam:quote), but it was getting messy and turning into a lot of work so I shelved it.

bricelam:
Woops, wrong issue. Ignore that

bricelam:
Discussed in https://github.com/aspnet/EntityFrameworkCore/issues/5581. May want to discuss again in the context of EF6

ericnewton76:
I did a whole pull request for it.  Fixed the unit tests which are doing string comparisons (problematic of course)

bricelam:
@ericnewton76 You rock; thanks for sending! We just need to discuss it as a team before merging. Don't forget that once it's merged, we the maintainers have to support this feature and fix any issues that customers find in it for the remaining lifetime of Entity Framework.

bricelam:
We agree that this sounds very useful when paired with `Add-Migration -Force`.

We don't like how much this implementation changes the main migration files. (e.g. You had to change a lot of tests to react to this change.) Is there a way to hide these partials in the `*.Designer.cs` migration file? This way you could still use the partials, but most users who don't need them won't have to see them every time the scaffold a new migration.

Another useful scenario is to be able to do something before/after every migration by simply changing the base type. This would require virtual methods that can be overridden by a user-defined base migration class.

Perhaps adding virtual methods instead of partial methods would make it a bit cleaner.

Thoughts?

ajcvickers:

PR #698, opened 2019-01-16
Remove EntityFramework.SqlServerCompact.Legacy

bricelam:
cc @ErikEJ


ErikEJ:
Everything has it's time!

PR #707, opened 2019-02-05
Fixed incorrect spelling of Debug statement in EntityEntry.cs

aslotte:
## Description
While debugging the code for another reason, I came across a minor spelling mistake, that I thought I could quickly correct.


divega:
Thanks @aslotte.

Issue #750, opened 2019-04-05
Flakey tests: Cannot open database "MigrationsTest"


bricelam:
Seeing occasional failures like this:

```
    [FATAL ERROR] System.Data.SqlClient.SqlException
      System.Data.SqlClient.SqlException : Cannot open database "MigrationsTest" requested by the login. The login failed.
      Login failed for user 'b00024U\vsagent'.
      Stack Trace:
           at System.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, UInt32 waitForMultipleObjectsTimeout, Boolean allowCreate, Boolean onlyOneCheckConnection, DbConnectionOptions userOptions, DbConnectionInternal& connection)
           at System.Data.ProviderBase.DbConnectionPool.TryGetConnection(DbConnection owningObject, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal& connection)
           at System.Data.ProviderBase.DbConnectionFactory.TryGetConnection(DbConnection owningConnection, TaskCompletionSource`1 retry, DbConnectionOptions userOptions, DbConnectionInternal oldConnection, DbConnectionInternal& connection)
           at System.Data.ProviderBase.DbConnectionInternal.TryOpenConnectionInternal(DbConnection outerConnection, DbConnectionFactory connectionFactory, TaskCompletionSource`1 retry, DbConnectionOptions userOptions)
           at System.Data.SqlClient.SqlConnection.TryOpenInner(TaskCompletionSource`1 retry)
           at System.Data.SqlClient.SqlConnection.TryOpen(TaskCompletionSource`1 retry)
           at System.Data.SqlClient.SqlConnection.Open()
        /_/test/EntityFramework/FunctionalTests/Migrations/TestHelpers/TestDatabase.cs(60,0): at System.Data.Entity.Migrations.TestDatabase.Execute[T](String commandText, Func`2 action, String connectionString)
        /_/test/EntityFramework/FunctionalTests/Migrations/TestHelpers/DatabaseProviderFixture.cs(66,0): at System.Data.Entity.Migrations.DatabaseProviderFixture.InitializeTestDatabase(DatabaseProvider provider, String databaseName)
        /_/test/EntityFramework/FunctionalTests/Migrations/TestHelpers/DatabaseProviderFixture.cs(24,0): at System.Data.Entity.Migrations.DatabaseProviderFixture..ctor()
```

Affects about nine tests.

PR #757, opened 2019-04-08
Update package branding to 6.3.0-preview5

bricelam:


bricelam:
Known failures (issue #750)

PR #771, opened 2019-04-17
Issue #561: Cast error, not enough information 

lizongshen:
Since our database table contains too many columns, in the production environment, after the cast error from the column to the model field, it is difficult to quickly locate the problem.

so I try to fix the issue #561 , look forward to your guidance and reply, thank you.


ajcvickers:

PR #774, opened 2019-04-17
Port tests to xUnit.net 2

bricelam:
This will enable them to cross-target .NET Core 3.0


test/EntityFramework/FunctionalTests/ProductivityApi/SimpleScenarios.cs
```diff
@@ -15,7 +15,6 @@ namespace ProductivityApiTests
     ///     They are here as a good starting point for understanding the functionality and to give some level
     ///     of confidence that the simple scenarios keep on working as the code evolves.
     /// </summary>
-    [PartialTrustFixture]
```
bricelam: @ajcvickers Is it OK to remove the partial trust tests or should I port them too?

ajcvickers: Remove them!

bricelam:
lol, I knew it would deadlock with all those Wait and Result calls.

roji:
Ah yes, xunit's deadlock detection feature :) https://github.com/xunit/xunit/issues/864

roji:
FYI just if it's too much work to fix the sync-over-async situation, you may be able to work around the issue via `-maxthreads unlimited`.

Issue #271, opened 2017-05-10
Make Entityframework 6 target .NET Standard and/or work with .NET Core


badre429:
On the build 2017 event @coolcsh(Scott hunter)  said that 70% in nuget library compatible with dotnet standard 2.0 is EF6 will support it.

divega:
@badre429 I will be repurposing your question to track the ongoing investigation.

jrkd:
"I would like to encourage anyone wanting this to happen to comment in the issue pointed above with their specific reasons. This will help us prioritize." - -divega 

Please, this would help me convert existing projects using EF6 to .net core.

douglaswaights:
We are running asp.net core on the full framework in 1.x and also plan to do this for the upcoming asp.net core 2.0 release. The reason for this is primarily we need EF6 for spatial. Maybe there are other blockers but spatial is the obvious one.

Would be great to run on .net core if we could, so modifying EF6 to work against net standard 2.x could be an interim solution until EF Core is ready ! :-)

Actually, we also need directory services and servicebase but these are flagged as high priority ports i believe, so hopefully later this year for them.

dimkanewtown:
Agree with @jrkd. Not having to migrate off EF6 would incredibly help migrate existing projects to dotnet core.

badre429:
@divega any update

marchy:
What is the current state of this gap after the .NET Core 2.0 release?

We are also stuck on EF6 for several reasons (in-memory query client-side eval, group by clauses, spatial etc.). We had to downgrade from EF7 to EF6 after a really painful migration and realizing we could not handle the present EF7 limitations after a few months – would be wonderful to be able to move to .NET Core while still targeting EF6.

Is this planned at all for EF 6.2 ??

ErikEJ:
API Port report here: 
[ApiPortAnalysis.xlsx](https://github.com/aspnet/EntityFramework6/files/1254381/ApiPortAnalysis.xlsx)

Rzpeg:
We are migrating our libraries to .NET Standard 2, and it would be great to have EF6 ported as well.

muehan:
Some Problem here. EF Core is no option for us. If EF6 does not work with .net Standard 2.0 I have to go back to the old Framework 4.x...

jayoungers:
Is the bulk of that 5% around the app.config piece?  If that was split out into a separate project how close would we be to standard v2?

fanoI:
[...]

PR #781, opened 2019-04-20
Port SimpleScenariosForLocalDb to .NET Core 3

bricelam:
Note, this is not a pattern for testing on .NET Core--just a quick and dirty way to get basic coverage for now.

Part of #271


test/EntityFramework/FunctionalTests.netcoreapp3.0/ProductivityApi/SimpleScenariosForLocalDb.cs
```diff
@@ -0,0 +1,427 @@
...
+    public class SimpleScenariosForLocalDb : FunctionalTestBase, IDisposable
+    {
+        #region Infrastructure/setup
+
+        private readonly object _previousDataDirectory;
+
+        public SimpleScenariosForLocalDb()
+        {
+            // TODO: Can we hardcode this somewhere like we do for the EF provider?
+            DbProviderFactories.RegisterFactory("System.Data.SqlClient", SqlClientFactory.Instance);
```
bricelam: All the test files were just coppied and trimmed to make them compile.

This is the only line added. Users will need to put this in their app startup for Preview 5.

test/EntityFramework/FunctionalTests.netcoreapp3.0/FunctionalTests.netcoreapp3.0.csproj
```diff
@@ -0,0 +1,25 @@
...
+    <PackageReference Include="xunit.assert" Version="$(XUnitVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\src\EntityFramework.SqlServer\EntityFramework.SqlServer.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <!-- HACK: xUnit.net uses AppDomains on .NET Framework to make App.config work -->
+    <None Update="xunit.console.dll.config">
```
bricelam: ☹️ This only makes `build.cmd` work. Test will fail in other test runners (including Visual Studio)

test/EntityFramework/FunctionalTests.netcoreapp3.0/ProductivityApi/SimpleScenariosForLocalDb.cs
divega: I guess we could add fallback logic every time we ask for the factory.

bricelam: I think we already have the indirection; I just didn’t want to rush the change. I’ll investigate Monday morning to see if there’s a simple change we can make to Preview 5.

test/EntityFramework/FunctionalTests.netcoreapp3.0/FunctionalTests.netcoreapp3.0.csproj
divega: What is the underlying issue? Is this related to the fact that a lot of functionality in EF6 still depends on entries in the config file? Assumig that is the case, will the tests work when/if we add sensible default behaviors? Or should the tests change to do code-based configuration?

divega: BTW, I am asking about what happens after preview5. I think tis is fine for now.

bricelam: I still need to investigate alternatives. There may be a way to change where ConfigurationManager looks...

PR #800, opened 2019-04-26
Don't simplify filter with custom functions over projection

kosinsky:
Don't simplify filter with custom functions over projection to avoid suboptimal SQL plans and allow window functions (these function not allowed in WHERE clause)

The change helps following cases:

- **Suboptimal SQL performance with custom functions**

For IQueryable that looks like:

```
context.Items.Select(i => new {i.Id, i.Title, Result = MyFuncs.MyFunc(i.Title)})
                      .Where(i => i.Result > 10)
```

EF6 simplifies expression and keeps only one SELECT with call to the function in SELECT and WHERE parts:
```
SELECT
       Id,
       Title,
       dbo.UDF_Test(Title) as L
FROM Items
WHERE dbo.UDF_Test(Title) > 10
```

Not simplified version should look like:

```
SELECT *
FROM (
     SELECT
           Id,
           Title,
           dbo.UDF_Test(Title) as L
     FROM Items
) T
WHERE L > 10
```
For table with 100K records. 
Simplified query execution takes: 1100 ms of SQL CPU time
Not simplified: 600 ms of SQL CPU time

If I replace dbo.UDF_Test by LEN (build in function), execution plans are the same and no performance difference are observed.

Custom UDF are block box and database have to call function twice for each row  and couldn't optimize it.

- **Window functions**
When instead of Custom UDF, window functions are used (ROW_NUMBER, LEAD, LAG, PERCENTILE_CONT etc.), SQL SERVER (and I assume every DB that supports window functions) will return error that window function supported only in SELECT or ORDER BY clauses


src/EntityFramework/Core/Query/PlanCompiler/TransformationRulesContext.cs
```diff
@@ -262,6 +262,49 @@ internal bool IsScalarOpTree(Node node, Dictionary<Var, int> varRefMap)
...
+        internal bool IncludeCustomFunctionOp(Node node, Dictionary<Var, Node> varMap)
+        {
+            PlanCompiler.Assert(varMap != null, "Null varRef map");
+
+            if (node.Op.OpType == OpType.VarRef)
+            {
+                var varRefOp = (VarRefOp)node.Op;
+                if (varMap.TryGetValue(varRefOp.Var, out var newNode))
+                {
+                    if (IncludeCustomFunctionOp(newNode, varMap))
```
divega: Collapse the whole `if`:

```suggestion
                    return IncludeCustomFunctionOp(newNode, varMap);
```

src/EntityFramework/Core/Query/PlanCompiler/TransformationRulesContext.cs
```diff
@@ -262,6 +262,49 @@ internal bool IsScalarOpTree(Node node, Dictionary<Var, int> varRefMap)
...
+                if (!functionOp.Function.BuiltInAttribute)
+                {
+                    return true;
+                }
+            }
+
+            // Simply process the result of the children.
+            for (var i = 0; i < node.Children.Count; i++)
+            {
+                if (IncludeCustomFunctionOp(node.Children[i], varMap))
```
divega: Again, just return the result of the call rather than having the `if`.

src/EntityFramework/Core/Query/PlanCompiler/FilterOpRules.cs
```diff
@@ -144,6 +145,12 @@ private static bool ProcessFilterOverProject(RuleProcessingContext context, Node
                 return false;
             }
 
+            // check to see that this predicate doesn't reference user-defined functions 
+            if (DbProviderServices.DisableFilterSimplificationForCustomFunctions && trc.IncludeCustomFunctionOp(predicateNode, varMap))
```
divega: Just in case this is not on purpose: here you are passing the varMap, which was computed for projectNode.Child1 bove. It is really enough or do we need to compute the var map over projectNode?

src/EntityFramework/Core/Common/DbProviderServices.cs
```diff
@@ -863,5 +866,16 @@ public virtual IEnumerable<object> GetServices(Type type, object key)
         {
             return _resolvers.GetServices(type, key);
         }
+
+        /// <summary>
+        /// By default expression like .Select(x => NewProperty = func(x.Property)).Where(x => x.NewProperty == ...) are simplified to avoid nested SELECT
+        /// In some cases, simplifing query with UDFs could caused to suboptimal plans due to calling UDF twice. Also some SQL functions aren't allow in WHERE clause.
+        /// Disabling that behavior
+        /// </summary>
+        public static bool DisableFilterSimplificationForCustomFunctions
```
divega: Name suggestion: `DisableFilterOverProjectionSimplificationForCustomFunctions`

src/EntityFramework/Core/Common/DbProviderServices.cs
```diff
@@ -863,5 +866,16 @@ public virtual IEnumerable<object> GetServices(Type type, object key)
         {
             return _resolvers.GetServices(type, key);
         }
+
+        /// <summary>
+        /// By default expression like .Select(x => NewProperty = func(x.Property)).Where(x => x.NewProperty == ...) are simplified to avoid nested SELECT
```
divega: Can't remember what the width limit is we are trying to follow in this codebase but this looks longer. Please word wrap.

src/EntityFramework/Core/Common/DbProviderServices.cs
divega: If the behavior was provider-specific, the provider's services class would be an ok option. It isn't really great, but at least there is precedent of us putting things there, and that is why I suggested it before.  

But now I understand that it isn't going to be provider specific. Having it on the base DbProvidderServices class feels very strange. I am wondering if we can add it as an instance flag on to DbConfiguration. There isn't much of a precedent (everything there is service based and not flags), but feels correct also from the perspective of:

- Query caching: you shouldn't be able to mutate the flag once EF is initialized. I can see it being convenient, but the behavior becomes unpredictable.
- DbConfiguration is our official code-based configuration story

@ajcvickers any thoughts?

src/EntityFramework/Core/Query/PlanCompiler/TransformationRulesContext.cs
divega: BTW, the my suggestion involves deleting a few subsequent lines but GitHub doesn't allow suggestions to do that yet.

src/EntityFramework/Core/Common/DbProviderServices.cs
kosinsky: @divega Did you mean DbContextConfiguration? I found that DbContextConfiguration has property UseDatabaseNullSemantics that passed down to PlanCompiler that uses it to decide that to apply/not apply some transformation rules (https://github.com/aspnet/EntityFramework6/blob/0ab1a198ee255ed7993b8657693d88bf8c4b7e77/src/EntityFramework/Core/Query/PlanCompiler/PlanCompiler.cs#L356) and Plan caching uses that property (it's renamed a bit) to decide that plan was cached or not (https://github.com/aspnet/EntityFramework6/blob/0ab1a198ee255ed7993b8657693d88bf8c4b7e77/src/EntityFramework/Core/Objects/ELinq/ELinqQueryState.cs#L101) It looks like pattern for transformation configuration

src/EntityFramework/Core/Query/PlanCompiler/FilterOpRules.cs
kosinsky: AFAIK, yes, it's enough.
Remapping that happens after new checks:
```
            //
            // Try to remap the predicate in terms of the definitions of the Vars
            //
            var remappedPredicateNode = trc.ReMap(predicateNode, varMap);
``` 
it looks in the predicate and rewrites all references from varMap by real expressions. My check ensures that undesired changes will not happen.

PS. I was thinking about implementing check inside ReMap, but that method ion general purpose and called in a lot of places. I'm not sure about performance impact of doing these checks over and over again.

src/EntityFramework/Core/Query/PlanCompiler/TransformationRulesContext.cs
kosinsky: Fixed.

src/EntityFramework/Core/Query/PlanCompiler/TransformationRulesContext.cs
kosinsky: We could have child without custom function before child with custom function. As a result we need to do Any(…), so I need if there, 

I modified tests to demo that case

src/EntityFramework/Core/Common/DbProviderServices.cs
kosinsky: I implemented the opt-in property at DbContextConfiguration. Please, take a look and let me know what do you think.

src/EntityFramework/Core/Common/DbProviderServices.cs
kosinsky: Done. However, it's in different fil now

src/EntityFramework/Properties/InternalsVisibleTo.cs
```diff
@@ -15,6 +15,10 @@
     InternalsVisibleTo(
         "EntityFramework.FunctionalTests.Transitional, PublicKey=002400000480000094000000060200000024000052534131000400000100010007D1FA57C4AED9F0A32E84AA0FAEFD0DE9E8FD6AEC8F87FB03766C834C99921EB23BE79AD9D5DCC1DD9AD236132102900B723CF980957FC4E177108FC607774F29E8320E92EA05ECE4E821C0A5EFE8F1645C4C0C93C1AB99285D622CAA652C1DFAD63D745D6F2DE5F17E5EAF0FC4963D261C8A12436518206DC093344D5AD293"
         )]
+[assembly:
+    InternalsVisibleTo(
+        "EntityFramework.FunctionalTests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007D1FA57C4AED9F0A32E84AA0FAEFD0DE9E8FD6AEC8F87FB03766C834C99921EB23BE79AD9D5DCC1DD9AD236132102900B723CF980957FC4E177108FC607774F29E8320E92EA05ECE4E821C0A5EFE8F1645C4C0C93C1AB99285D622CAA652C1DFAD63D745D6F2DE5F17E5EAF0FC4963D261C8A12436518206DC093344D5AD293"
+        )]
```
ajcvickers: @bricelam Do you think it's okay to add `InternalsVisibleTo` here given where we are in the development cycle for EF6?

kosinsky: I needed it to disable plan caching it tests, but after I addressed Diego's feedback don't need it. I've reverted that change.

ajcvickers: Thanks!

ajcvickers:
Thanks @kosinsky. Sorry for the long delay.

divega:
+1 thanks for this @kosinsky!

Issue #805, opened 2019-04-29
Change YAML pipeline to upload package artifacts only in successful builds


dougbu:
Without this RaaS gests confused and reports the package upload failure as the main problem e.g. build https://dnceng.visualstudio.com/public/_build/results?buildId=171963 and report https://github.com/aspnet/AspNetCore-Internal/issues/2413

dougbu:
3f1f7570d607

PR #806, opened 2019-04-29
Publish package artifacts only in successful builds

dougbu:
- #805


dougbu:
https://dev.azure.com/dnceng/public/_build/results?buildId=173202 nicely demonstrated the fix. Tests failed and the package artifact upload step was skipped. Pushing to 'master' 😺

Issue #823, opened 2019-05-07
Consider supporting Microsoft.Data.SqlClient in EF6


divega:
Currently EF6 depends on System.Data.SqlClient which is included in .NET Framework, but SQL Server is moving development to Microsoft.Data.SqlClient, where any new features will be supported. 

EF6 is in a different situation from EF Core, in that we don't have any breaking change release ahead of us, so I think we probably shouldn't change what provider we use by default, but it should be easy to enable the new one to work.

ajcvickers:
@divega I was just thinking I needed to file this! :-)

What do you think about depending on the new package (and only the new package) for .NET Core, and keeping the old dependency for .NET Framework?

/cc @bricelam

divega:
@ajcvickers, from the perspective that the combination of EF6 + .NET Core is completely new and has no compat baggage per se, it seems alright. But it feels a bit weird to diverge so much across TFMs, plus in principle I think we should support the new provider in .NET Framework anyway (I assume in a new package), so not sure it buys us much in terms of simplicity. But this is just my first take on it. I am happy to discuss it.

bricelam:
Will this make porting apps to .NET Core easier or harder?

ErikEJ:
My vote is stay with System.Data.SqlClient - it will receive no new features, just like EF6 🤡

bricelam:
I think the tension is between:
* Having to change all your using statements, or
* Missing features that you were using on .NET Framework

ErikEJ:
What missing features (other than Always Encrypted) ?

divega:
For now, nothing but Always Encrypted, I think. But that one alone is a pretty big deal.

ajcvickers:
We decided that the existing provider will continue to use `System.Data.SqlClient`, which is non-breaking. However, we will likely release a new provider that uses `Microsoft.Data.SqlClient` at some point post 6.3, since it closes gaps important gaps on .NET Core, such as Always Encrypted.

jtlimson:
+1

CZEMacLeod:
@ajcvickers Does this mean that the existing SqlClient provider will be shipped as a separate NuGet package (as a dependency?) and allow you to use it or the new MS.D.SC or will we be landed with having to load both providers / assemblies into memory?
I always found it slightly odd that even if you were using sqlite or another provider, you ended up with the SqlServer provider anyway.

ajcvickers:
[...]

PR #849, opened 2019-05-16
Port FunctionalTests.Transitional to .NET Core 3

bricelam:
Part of #271


src/EntityFramework/Infrastructure/DependencyResolution/DefaultProviderFactoryResolver.cs
```diff
@@ -31,6 +32,11 @@ private static object GetService(Type type, object key, Func<ArgumentException,
                 }
                 catch (ArgumentException e)
                 {
+                    if (string.Equals(name, "System.Data.SqlClient", StringComparison.OrdinalIgnoreCase))
+                    {
+                        return SqlClientFactory.Instance;
```
bricelam: @ajcvickers @divega If we don't find a factory for SqlClient, we'll return this hard-coded one.

src/EntityFramework/Utilities/DbProviderFactoryExtensions.cs
```diff
@@ -29,6 +30,11 @@ public static string GetProviderInvariantName(this DbProviderFactory factory)
 
             if (row == null)
             {
+                if (factory is SqlClientFactory)
+                {
+                    return "System.Data.SqlClient";
```
bricelam: Ditto for mapping the factory to its invariant name.

src/EntityFramework/Infrastructure/DependencyResolution/DefaultProviderFactoryResolver.cs
divega: I think this is fine. It should fix the usability issues, but I am not sure about the full implications. Maybe check it in but then talk about it a bit in the design meeting. It also has some implications for  https://github.com/aspnet/EntityFramework6/issues/823 (e.g. using Microsoft.Data.SqlClient).

src/EntityFramework/Utilities/DbProviderFactoryExtensions.cs
divega: Same as above.

src/EntityFramework/Infrastructure/DependencyResolution/DefaultProviderFactoryResolver.cs
ajcvickers: Agreed that this is fine but we need to talk about #823. I have some ideas.

PR #852, opened 2019-05-18
Enable building on macOS and Linux

bricelam:


eng/Versions.props
```diff
@@ -5,6 +5,7 @@
     <SemanticVersioningV1>True</SemanticVersioningV1>
     <IncludeSourceRevisionInInformationalVersion>False</IncludeSourceRevisionInInformationalVersion>
     <UsingToolNetFrameworkReferenceAssemblies>True</UsingToolNetFrameworkReferenceAssemblies>
+    <MicrosoftNetFrameworkReferenceAssembliesVersion>1.0.0-preview.1</MicrosoftNetFrameworkReferenceAssembliesVersion>
```
bricelam: Looks like this version still doesn't have the fix for dsplaisted/ReferenceAssemblyPackages#2

cc @dsplaisted

bricelam: cc @NikolaMilosavljevic

dsplaisted: We have this tracked at https://github.com/dotnet/core-sdk/issues/1912.  It should be fixed soon.

bricelam: Thanks!

dsplaisted: We just published an update that should fix this: https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/1.0.0-preview.2

bricelam:
Everything appears to be working as expected now. Development on macOS and Linux is unblocked.

We're not going to enable macOS and Ubuntu on the CI since there are still 4K tests failing with "LocalDB is not supported on this platform." If the community is interested in skipping these tests, or getting the base connection string from an environment variable, we'd happily accept a PR.

PR #864, opened 2019-05-21
Port FunctionalTests.ProviderAgnostic to .NET Core 3

bricelam:
Part of #271


test/EntityFramework/FunctionalTests.ProviderAgnostic/FunctionalTests.ProviderAgnostic.csproj
```diff
@@ -38,4 +38,11 @@
     <ProjectReference Include="..\FunctionalTests.Transitional\FunctionalTests.Transitional.csproj" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
+    <None Update="App.config">
+      <Link>xunit.console.dll.config</Link>
```
bricelam: @ajcvickers This is the only way to get App.config in xUnit.net on .NET Core. I can't use DbConfiguration because there's no way to register one for a context in a different assembly. The only way to do that is using App.config.

ajcvickers: It is what it is...

PR #892, opened 2019-06-04
Add Microsoft.Data.Entity.Build.Tasks

### LINKED ISSUE #828: EDMX Metadata Generator options for .NET Core
Per the announcement post, EF 6.3 Preview 5 is currently shipping without EDMX support. Fortunately, I solved that problem about a year ago for SDK-style projects targeting the .NET Framework. I've open-sourced that tool here: https://github.com/CloudNimble/EF6.3MetadataGenerator.

Elements of this tool could be turned into a .NET Core Global Tool using `dotnet-script` and csx files (or possibly another option), but the code would have to be changed to manipulate the Project XML manually.

[According to this tweet](https://twitter.com/divega/status/1126271044505853952), the team is considering a couple different options. I thought this thread would be a good place to discuss those options, and consider if the tool I linked to would be useful in the final product.

bricelam:
*@bricelam seven years ago:*
We don't need to open source that. It never changes and will always ship as part of .NET Framework.

*@bricelam seven years later:*
Crap. We need it on .NET Core.

Changes to the original code:
- Updated file headers
- Renamed BuildTasksResources.resx to Properties/Resources.resx
- Removed unused XmlConstants members
- Simplified EntityDesignerUtils by inlining some constants

Fixes #828


bricelam:
@natemcmaster @pranavkm Any gotchas implementing a cross-platform MSBuild task or using NuGet's `buildTransitive` directory?

pranavkm:
@javiercn used the `buildTransitive` feature recently with build targets, he might know more.

javiercn:
AFAIK it works just fine.

javiercn:
The way we build packages is we create three folders and have buildTransitive point to buildMultitargetting and that one point to build

bricelam:
Oh interesting. Presumably for compatibility with different NuGet clients? I think this one is TFM-specific so buildCrosstargeting may not make sense, but `build` might.

bricelam:
Yeah, i think I’ll add a `build` directory. This would enable net40 projects building with older NuGet clients.

natemcmaster:
@javiercn's suggestion is good one. Only one of the `build*/` folder conventions is used at a time. I've adopted the practice of only shipping one copy of the "real" props and targets which understand the MSbuild context they are in (inner or outer build), and then adding several dummy targets which match the various NuGet conventions, build/, buildMultiTargeting/, buildTransitive/, sdk/, etc.

Another gotcha is that there is no way to make `buildMultiTargeting/`, but since it doesn't appear you need to import props/targets in the outer build of a multi-targeting app, you're okay on this one.

/cc @jainaashish in case he is aware of any other gotchas with `buildTransitive/`

bricelam:
Updated in https://github.com/aspnet/EntityFramework6/pull/894

PR #893, opened 2019-06-04
Update the Signing Certificate

lajones:
Previous cert was SHA1 and SHA256. Going forward MicroBuild will support SHA256 exclusively. So changing cert.


bricelam:
LGTM. That's the one the runtime uses via Arcade.

Issue #189, opened 2017-02-02
Optimize EF6 to work with PackageReference and no app.config


ErikEJ:
(Edited by @divega. Re-purposing this issue to detail the current action plan on this area. The original text of the issue was kept at the bottom)

We had a discussion with the NuGet team on what we can do to adapt the EF6 package to a world in which more customers use NuGet's PackageReference and Install.ps1 doesn't execute anymore and in which potentially package.config is discouraged or even disabled in the long term. Here are a few things we want to make sure are enabled:

- [x] **No breaking change:** Application with existing EF6 entries in app.config should continue to work based on the existing configuration when upgrading
- [x] **F5 experience works:** New applications that download the EF6 package without app.config should be able to work out of the box with (e.g. F5 experience) SQL Server, likely assuming (localdb)\mssqllocaldb instance. This may required changes to default provider and connection factory registration   
- [ ] **Works for other providers:** New applications that download an EF6-based provider package should be able to work easily 
  - [ ] For this to happen automatically we may need a new mechanism to configure the provider, e.g. could we use MSBuild props or targets as detailed at https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#including-msbuild-props-and-targets-in-a-package?
  - [ ] Otherwise having good coverage in the documentation on how to do this manually (which would require finishing [porting EF6 to docs.microsoft.com](https://github.com/aspnet/EntityFramework.Docs/pull/439)) might be acceptable
  - [ ] We need to identify existing providers that use the recommended PowerShell commands to register the provider and understand in what state the changes would leave them
- [ ] ~~**Commands advertising:** Since install.ps1 won't run, we won't be able to display a banner telling customers about the available commands. Should we resort to a readme.txt file instead? This actually affects EFCore tools as well~~ (Not needed)
- [ ] ~~**T4 template distribution:** Currently we use NuGet packages (entityframework.codetemplates.csharp and entityframework.codetemplates.visualbasic) to distribute customizable versions of the T4 templates for the "code first from an existing database" option in the EF6 designer. We would need to have alternative ways to distribute them, e.g. add them as item templates in the designer's setup, or point customers to the code in docs, github, etc.~~ (Moved to #883)

@bricelam @ajcvickers please add any details I missed.

---
# [VS2017RC3] No app.config created when using "packageReference" iso packages.config
File, New Project, Console Appliation, Target .NET 4.0, add EntityFramework package, use PackageReference,

Result: app.config file with EF settings created

[...]

PR #895, opened 2019-06-04
Default to (localdb)\MSSQLLocalDB

bricelam:
Part of #189

### DIFF
--- a/src/EntityFramework/Infrastructure/LocalDbConnectionFactory.cs
+++ b/src/EntityFramework/Infrastructure/LocalDbConnectionFactory.cs
@@ -26,6 +26,14 @@ namespace System.Data.Entity.Infrastructure
         private readonly string _localDbVersion;
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="LocalDbConnectionFactory"/> class.
+        /// </summary>
+        public LocalDbConnectionFactory()
+            : this("mssqllocaldb")
+        {
+        }
+
+        /// <summary>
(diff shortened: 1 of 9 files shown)


src/EntityFramework/Infrastructure/LocalDbConnectionFactory.cs
```diff
@@ -25,6 +25,14 @@ public sealed class LocalDbConnectionFactory : IDbConnectionFactory
         private readonly string _baseConnectionString;
         private readonly string _localDbVersion;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LocalDbConnectionFactory"/> class.
+        /// </summary>
+        public LocalDbConnectionFactory()
+            : this("mssqllocaldb")
```
ajcvickers: Better hope they don't change it again!

ajcvickers:
@divega @bricelam Do you think we should announce this as a breaking change?

src/EntityFramework/Infrastructure/LocalDbConnectionFactory.cs
ErikEJ: It's been the same for 4 major SQL Server versions now

divega:
@ajcvickers Maybe we should start putting together a What's new page for EF 6.3 in the docs and put it there. But I am fine with announcing it as a breaking changes via other means as well.

ajcvickers:
Yes, the What's New page sounds good. Let's start with that.

PR #909, opened 2019-06-10
Increase build timeout

JunTaoLuo:
Seems like the builds timeout after 2.5 hours see https://github.com/aspnet/AspNetCore-Internal/issues/2652. The successful builds in this repo often takes almost that long so we should just increase the limit.


JunTaoLuo:
This is a pretty trivial change, so I'll merge this without waiting for the build checkmark.

bricelam:
Thanks! I forgot to increase this after adding netcoreapp3.0 to all the tests.

PR #910, opened 2019-06-10
Change dual SHA1-SHA256 signing to SHA256-only (VS2017)

lajones:
Updating VS2017 branch to change dual SHA1-SHA256 signing to SHA256-only.


lajones:
@bricelam Same change as #893 - but for the VS2017 branch.

Issue #788, opened 2019-04-22
Browser Model not visible


ajcvickers:
From VS Feedback:

The Model Browser window & #232; empty or not visible

ajcvickers:
**EF Team Triage:** This issue is lacking enough information for us to be able to effectively triage it. In particular, it is missing the following information requested in the new issue template. Can you please provide this information?

> ### Steps to reproduce
Ideally include a complete code listing that we can run to reproduce the issue.
Alternatively, you can provide a project/solution that we can run.

BTW we're not just doing this to be mean :smile:... we get a lot traffic on this project and it takes time to attempt to reproduce an issue based on fragments of information. In addition, our attempt is often unsuccessful as the exact conditions required to hit the issue are often not explicitly included in the code provided. To ensure we maximize the time we have to work on fixing bugs, implementing new features, etc. we ask that folks give us a self-contained way to reproduce an issue.

For a guide on submitting good bug reports, read [Painless Bug Tracking](http://www.joelonsoftware.com/articles/fog0000000029.html). 

*BTW this is a canned response and may have info or details that do not directly apply to this particular issue. While we'd like to spend the time to uniquely address every incoming issue, we get a lot traffic on the EF projects and that is not practical. To ensure we maximize the time we have to work on fixing bugs, implementing new features, etc. we use canned responses for common triage decisions.*

lstauber:
Same here. 16.1, ef model, model browser empty or containing data from previously active tab. Steps to reproduce - open edmx designer with old edmx file and then model browser.

melcug:
I confirm the issue.
I'm using Visual Studio 2019 version 16.1.1 and EF version 6.2.
In order to reproduce this issue, I created a model based on an existing database, the application is built it's running with no errors.
But the Entity Data Model Browser Window is empty or gets "parts" from other opened windows (see bellow).
![EF Model Browserl](https://user-images.githubusercontent.com/8708178/58753148-d8111c80-84c3-11e9-9a22-22bf2cf49bd5.png)

roblarky:
Same issue here. This is a brand new laptop, so it has only ever had VS 2019 Community installed, no prior versions. The solution was created in VS 2017. The Model Browser panel renders fine for the same solution on my other systems in VS 2019, though those machines also have VS 2017 installed, in case that helps any.

[...]

PR #934, opened 2019-06-20
Use DpiAwarenessContext.SystemAware when creating Tool Windows.

lajones:
Fix for internal bug 898367 and #788 - Model Browser not visible.

This only happens (for me) on some screens, and then only for some combinations of screen resolution and scale factor. It could be worked around by unchecking **Tools -> Options -> Environment -> General -> Optimize rendering for screens with different pixel densities**, but this fix should make that unnecessary.

### DIFF
--- a/src/EFTools/EntityDesign/CodeGeneration/Generators/TextTemplatingHost.cs
+++ b/src/EFTools/EntityDesign/CodeGeneration/Generators/TextTemplatingHost.cs
@@ -4,6 +4,7 @@
     using System.CodeDom.Compiler;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
     using System.IO;
     using System.Linq;
     using System.Text;
@@ -109,6 +110,7 @@
             return new TextTemplatingSession();
         }
 
+        [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope")]
         public string ProcessTemplate(string inputFile, string content, ITextTemplatingCallback callback = null)
         {
             Debug.Assert(!string.IsNullOrEmpty(inputFile), "inputFile is null or empty.");
--- a/src/EFTools/EntityDesignPackage/EntityDesignPackage.csproj
+++ b/src/EFTools/EntityDesignPackage/EntityDesignPackage.csproj
@@ -55,6 +55,9 @@
   </PropertyGroup>
   <Import Project="$(RepositoryRoot)tools\EFToolsVsReferences.settings.targets" />
   <ItemGroup>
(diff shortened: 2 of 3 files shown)


src/EFTools/EntityDesignPackage/EntityDesignPackage.csproj
```diff
@@ -66,11 +69,15 @@
     <Reference Include="Microsoft.VisualStudio.Shell.Interop.11.0" />
     <Reference Condition="'$(VisualStudioVersion)' != '11.0' AND '$(VisualStudioVersion)' != '12.0' AND '$(VisualStudioVersion)' != '14.0'"
                Include="Microsoft.VisualStudio.Shell.Interop.15.3.DesignTime" />
+    <Reference Condition="'$(VisualStudioVersion)' == '16.0'" Include="Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime, Version=14.0.0.0">
```
divega: I am a bit surprised that we pull the 14.0 version of the assemblies for the 16.0 version of VS? Shouldn't the major version numbers match?  

Also, is it ok to have Microsoft.VisualStudio.Shell.Interop.15.3.DesignTime (per the previous line) and Microsoft.VisualStudio.Shell.Interop.14.0.DesignTime added as references at the same time? It seems both conditions apply if the version of VS is 16.0.

src/EFTools/EntityDesign/CodeGeneration/Generators/TextTemplatingHost.cs
```diff
@@ -109,6 +110,7 @@ public ITextTemplatingSession CreateSession()
             return new TextTemplatingSession();
         }
 
+        [SuppressMessage("Microsoft.Reliability", "CA2000:Dispose objects before losing scope")]
```
divega: Curious, what object is disposable here? Engine? Asking because the warning may be flagging an actual bug in our code.

src/EFTools/EntityDesignPackage/EntityDesignPackage.csproj
lajones: The 14.0 version is required in the fix published by the PMA team for this (which is specifically for 16.0 VS onwards). I agree that 15.3 and 14.0 is weird but does not appear to cause any problems and taking out 15.3 would require extensive re-testing.

src/EFTools/EntityDesign/CodeGeneration/Generators/TextTemplatingHost.cs
lajones: Yes, `Engine`. That code has been like that for multiple major versions - so it's stable and I'd rather not destabilize it while trying to fix this urgent issue. Plus we have many suppressions like this - so it's a more general task to go through the code and remove them.

src/EFTools/EntityDesignPackage/EntityDesignPackage.csproj
lajones: And note, 15.3 by itself did *not* work.

PR #935, opened 2019-06-23
Move to 3.0 Preview 6 SDK

dougbu:


eng/Versions.props
```diff
@@ -38,4 +38,11 @@
     -->
     <MicrosoftNETCoreAppRuntimeVersion>$(MicrosoftNETCoreAppRuntimewinx64PackageVersion)</MicrosoftNETCoreAppRuntimeVersion>
   </PropertyGroup>
+  <!-- Restore feeds -->
+  <PropertyGroup Label="Restore feeds">
+    <RestoreSources Condition=" '$(DotNetBuildOffline)' != 'true' ">
```
bricelam: Is this for source build? We don't have to do that for EF6, do we?

dougbu: Source build isn't a direct issue for EF 6 or EF Core but it is an issue to either (a) leave our builds behind on older SDKs or (b) get to far ahead of where AspNetCore is in EF Core.

dougbu: And, these extra restore sources are necessary only because the Preview 6 SDK no longer implicitly adds our internal feeds.

bricelam: But the condition on this line is just dead code, right?

dougbu: No, it's the pattern we need to put in place for servicing builds et cetera.

Issue #231, opened 2017-03-24
Make EF NuGet package / migrations and EF Tools work with new csproj format


ErikEJ:
**Workaound:**

Currently the following workaround works (tested on an ASP.NET Core 2.0 app):
- Move the model into a traditional .NET Framework library that uses the old project system
- Add the aforementioned library as a dependency
- Unload the ASP.NET Core application from the solution
- Execute the Enable-migrations command against the library project. This step seems to only be necessary the first time, otherwise you will get an error saying: 
   > Exception calling "SetData" with "2" argument (s): "Type 'Microsoft.VisualStudio.ProjectSystem.VS.Implementation.Package.Automation.OAProject' in assembly 'Microsoft.VisualStudio.ProjectSystem.VS.Implementation, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' is not marked as serializable."
- Re-load the ASP.NET Core project
- Work as normal, with migration commands always pointing to the .NET Framework library project

**Options:**

We could consider
1. Doing the full work to enable EF6 migrations to work with the new project system (which would likely be very expensive)
2. Improve EF6 migrations to handle the presence of projects using the new project system more gracefully, e.g.:
   1. It would be nice if we provided a nice error which provided guidance on creating a separate project
   2. Once a customer applies the workaround, it would be nice if they didn't need to unload the ASP.NET Core project

**Original issue:**

See https://github.com/aspnet/EntityFramework6/issues/230 and https://github.com/aspnet/EntityFramework6/issues/189 and https://github.com/NuGet/Home/issues/4876 

See also _New csproj format and GenerateSeparateFiles=true causes VS2017 to freeze  https://github.com/sjh37/EntityFramework-Reverse-POCO-Code-First-Generator/issues/266_ which blocks EF Tools tt codegen with the new csproj files.

lincolnzocateli:
Hi, 
I am with a system that has migrated from VS2015 (net461) to VS2017 (csproj "net461"), with problem equal ao @HellBrick.#230 

Do you have prevision to new version EF 6.3.2 ?
Is there any workaround I can do for now?

weitzhandler:
I'm also having this issue in EF 6.1.3 in a WebAPI + Xamarin PCL.
Unable to execute `PM> Enable-Migrations -ProjectName MyWebAPI`.

I get the following exception:

PM> enable-migrations -projectname mywebapiprojectname
[...]

PR #953, opened 2019-06-27
Don't call Add-EFProvider or Add-EFDefaultConnectionFactory

bricelam:
Part of #231

### DIFF
--- a/src/EntityFramework.SqlServerCompact/EntityFramework.SqlServerCompact.nuspec
+++ b/src/EntityFramework.SqlServerCompact/EntityFramework.SqlServerCompact.nuspec
@@ -16,6 +16,5 @@
     <file src="../../artifacts/bin/EntityFramework.SqlServerCompact/$Configuration$/net45/EntityFramework.SqlServerCompact.dll" target="lib/net45/" />
     <file src="../../artifacts/bin/EntityFramework.SqlServerCompact/$Configuration$/net45/EntityFramework.SqlServerCompact.xml" target="lib/net45/" />
     <file src="content\**\*" target="content/" />
-    <file src="tools\**\*" target="tools/" />
   </files>
 </package>
--- a/src/EntityFramework.SqlServerCompact/content/net40/App.config.install.xdt
+++ b/src/EntityFramework.SqlServerCompact/content/net40/App.config.install.xdt
@@ -1,0 +1,13 @@
+﻿<configuration xmlns:xdt="http://schemas.microsoft.com/XML-Document-Transform">
+  <entityFramework>
+    <defaultConnectionFactory xdt:Transform="Remove" />
+    <defaultConnectionFactory type="System.Data.Entity.Infrastructure.SqlCeConnectionFactory, EntityFramework" xdt:Transform="Insert">
+      <parameters>
+        <parameter value="System.Data.SqlServerCe.4.0" />
+      </parameters>
+    </defaultConnectionFactory>
+    <providers>
+      <provider invariantName="System.Data.SqlServerCe.4.0" type="System.Data.Entity.SqlServerCompact.SqlCeProviderServices, EntityFramework.SqlServerCompact" xdt:Transform="InsertIfMissing" xdt:Locator="Match(invariantName)" />
+    </providers>
+  </entityFramework>
+</configuration>
--- a/src/EntityFramework.SqlServerCompact/content/net40/Web.config.install.xdt
+++ b/src/EntityFramework.SqlServerCompact/content/net40/Web.config.install.xdt
@@ -1,0 +1,13 @@
+﻿<configuration xmlns:xdt="http://schemas.microsoft.com/XML-Document-Transform">
+  <entityFramework>
+    <defaultConnectionFactory xdt:Transform="Remove" />
+    <defaultConnectionFactory type="System.Data.Entity.Infrastructure.SqlCeConnectionFactory, EntityFramework" xdt:Transform="Insert">
+      <parameters>
+        <parameter value="System.Data.SqlServerCe.4.0" />
+      </parameters>
+    </defaultConnectionFactory>
+    <providers>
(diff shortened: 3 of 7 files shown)


bricelam:
Should we remove Add-EFProvider? (We still need Add-EFDefaultConnectionFactory to dynamically choose between SQL Express and LocalDB)

The only providers that I can see using it are Firebird, Npgsql, and SQLite. They could update to this pattern.

cc @cincuranet @roji @mistachkin

divega:
What is the experience if you install one of these existing providers on top of the EF 6.3 package and those scripts are missing? 

Is our own old SQLCE provider using it?

divega:
> Is our own old SQLCE provider using it?

Ah, I asked this before I looked at the changes :smile:

> We still need Add-EFDefaultConnectionFactory to dynamically choose between SQL Express and LocalDB

Is there a way to use these transforms so that on package upgrade we don't touch the defaultConnectionFactory if there is one there? Or does the uninstall transform always need to run followed by the install transform on package upgrade? 

Asking because I think ideally we would stop using Add-EFDefaultConnectionFactory.

bricelam:
I realized we can internalize the logic for Add-EFDefaultConnectionFactory and remove it too.

bricelam:
> What is the experience of these scripts are missing

All of this only works with packages.config so PackageReference is unaffected.

We could make them warm and no-op, but I’m not sure you can even get into that situation. There would have to be an update to the provider package that still used the cmdlets.

bricelam:
Or you’d have to install the EF 6.3 package *then* an older provider. Don’t we tell customers to just install the provider package?

divega:
> All of this only works with packages.config so PackageReference is unaffected.

Yeah, I was reading about that just now. Thanks for explaining.

> We could make them warn and no-op, but I’m not sure you can even get into that situation. There would have to be an update to the provider package that still used the cmdlets.

Ok, I might be missing it completely, but what I had in mind, is to install EF 6.3 and the install a provider that hasn't been updated and still relies on the cmdlets (but that it is otherwise compatible with EF 6.3 because we are not making breaking changes).

> I realized we can internalize the logic for Add-EFDefaultConnectionFactory and remove it too.

Crazy idea (sorry for the randomization if it is too crazy): what if EF 6.3 didn't do any transform on install (even for package.config) but also didn't do any transform on uninstall? 

I am speculating that this would result in:

1. Upgrades from previous package versions preserve whatever configuration was in place and hence continue to work the same way.
2. Installation of the 6.3 package in clean projects would just defer decisions to runtime. 

Are we at a point in which the runtime (and providers) can choose reasonable defaults when there is zero configuration? 

A couple of pieces I imagine could be missing are:
1. Discover the provider assembly
2. Choose a default connection factory different from LocaldDB if that one is also available. I assume that could be a problem always if installing providers always results in the SQL Server provider also being installed. For the cases in which there are legitimately multiple providers, I think requiring some additional code or configuration to choose the default is reasonable.

divega:
> Or you’d have to install the EF 6.3 package then an older provider. Don’t we tell customers to just install the provider package?

I guess yes, it is not unreasonable to do this if you want some new feature from EF 6.3 but the provider hasn't been updated.

bricelam:
(Updated to revert changes on uninstall)

bricelam:
> Why are we switching to the transformation files?

This PR was mostly to prove that providers can preserve existing behavior without Add-EFProvider and Add-EFDefaultConnectionFactory. I'd prefer to remove them rather than port them to the new architecture.

ajcvickers:
I know we tried transforms for this a long time ago and the experience wasn't great, but I don't remember the details. However, I think it's probably good enough anyway.

I'm going to approve this, but we can still discuss other options.

bricelam:
Discussed with @divega, and we're going to keep Add-EFProvider & Add-EFDefaultConnectionFactory

bricelam:
@ajcvickers This is using the "new" (since obsoleted) XDT transforms which are more powerful than `app.config.transform` which was the only thing available when we first tried.

bricelam:
TL;DR for @cincuranet, @roji & @mistachkin:

The pattern of calling `Add-EFProvider` inside `init.ps1` will continue to work. Optionally, you can update to the pattern shown in this PR using `app.config.install.xdt` instead.

cincuranet:
@bricelam Thanks. 

I'm planning to update to XDT ([ticket](http://tracker.firebirdsql.org/browse/DNET-883)).

roji:
@bricelam any tips on how to get those four files into the nupkg correctly from a project that only has a csproj (am tired from fighting with nuget)?

```xml
  <ItemGroup>
    <Content Include="content/**" PackagePath="contentFiles/any">
      <IncludeInPackage>true</IncludeInPackage>
    </Content>
  </ItemGroup>
```

This gets the files in the nupkg, but in the nuspec they're listed under `<contentFiles>` which seems to make nuget simply copy them (rather than apply transformations).

Thanks...

bricelam:
@roji They need to go under `content`. It only works with packages.config (not PackageReference)

cincuranet:
@roji Any reason you included also the `Xxx.config.transform files` in i.e. https://github.com/roji/EntityFramework6.Npgsql/commit/717e7fa8031c30f8e7de915cceb29e3079001e16?

roji:
@cincuranet I honestly don't remember any more... I think maybe I just followed the example from EF6?

cincuranet:
@roji I now tested only with XDT and everything works correctly (for packages.config). Probably good for me. 🤷

@bricelam Any ideas for PackageReference (for the future maybe)?

bricelam:
@cincuranet Issue #920 will help with the ADO.NET part of things. Does anyone actually use the default connection factory? I imagine most people use a provider-specific connection object or a connection string from config that also specifies the provider.

cincuranet:
Never seen the default connection factory to be used in real code.

PR #954, opened 2019-06-28
Simplify Add-EFProvider & Add-EFDefaultConnectionFactory

bricelam:
Part of #231


src/NuGet/EntityFramework/tools/EntityFramework.psm1
```diff
@@ -10,159 +10,14 @@ $knownExceptions = @(
...
-            }
-
-            throw $error.Message
-        }
-    }
-    finally
-    {				
-        Remove-Runner $runner
-    }
+    Write-Warning 'Add-EFDefaultConnectionFactory is obsolete. Providers should use an XDT transform instead.'
```
bricelam: We could just as easily keep these working by using the code below.

bricelam: Going to keep these

bricelam:
(Manually tested using the current version of System.Data.SQLite.EF6 which calls Add-EFProvider)

Issue #917, opened 2019-06-12
Error in generate database from model using EF6 tools


divega:
### Repro steps: 
1. Create an empty .NET Core console app
2. Add EF 6.3 preview 6 to it 
(Note: I suspect the two previous steps can be skipped, but I haven't verified)
3. Create an empty .NET Framework console app
4. Created an empty EDMX file using the wizard
5. Add an entity type on the designer
6. Right-click on the designer surface and click on the _Generate database from model_ option
7. Create a new connection to SQL Server LocalDB (first step of the wizard)
8. Select EF 6.x (second step of the wizard)
9. Click on the next button

### Result:
An error dialog is displayed containing a stack trace. The text starts with:
> Encountered the following errors while processing the template 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\Common7\IDE\Extensions\Microsoft\Entity Framework Tools\DBGen\SSDLToSQL10.tt: Line 0, Column 0: 'An exception was thrown while trying to compile the transformation code. Teh following Exception was thrown: System.IO.FileNotFoundException: Could not find the file 'C:\Program files (x86)\Microsoft Visual Studio\2019\Preview\Common7\IDE\EntityFramework.SqlServerCompact.dll'...

### Screen shot:
![image](https://user-images.githubusercontent.com/1253414/59381415-2ae2a380-8d10-11e9-9af1-f9323c3902b3.png)

ajcvickers:
@divega What VS version?
@lajones Could this be a result of moving the templates off the `C:\` drive?

divega:
I was able to repro on my laptop with 

Microsoft Visual Studio Enterprise 2019 Int Preview
Version 16.2.0 Preview 2.0 [29001.23.d16.2stg]
VisualStudio.16.IntPreview/16.2.0-pre.2.0+29001.23.d16.2stg
Microsoft .NET Framework
Version 4.8.03752

I will try to remember checking what I had in my desktop when I found it earlier.

divega:
Repros on this as well:

Microsoft Visual Studio Enterprise 2019 Int Preview
Version 16.2.0 Preview 3.0 [29011.400.d16.2stg]
VisualStudio.16.IntPreview/16.2.0-pre.3.0+29011.400.d16.2stg
Microsoft .NET Framework
Version 4.8.03752

lajones:
[...]

PR #956, opened 2019-06-29
Re-include EntityFramework.SqlServerCompact in setup

lajones:
Updated the way MicroBuild works to use their version of NuGet (with their credentials manager) to restore the packages needed for setup. Then keep only the part of the following MicroBuild task which extracts the nuspecs. This allows us to re-include EntityFramework.SqlServerCompact in setup.

Addresses issue #917.


lajones:
Fix will be in 16.2 Preview 4.

PR #958, opened 2019-06-29
fix: CREF issues from Docs Repo

nschonni:
Ported most of the cref changes from https://github.com/aspnet/EntityFramework.ApiDocs/pull/58 that are being flagged by the OpenBuild system


ajcvickers:
Thanks!

PR #971, opened 2019-07-02
Query compilation performance - Part 2

chrfin:
This is an update to #430 with the addition to be able to configure or completely disable the query optimizations as already discussed in the initial pull request.


src/EntityFramework/Utilities/IntegerSwitch.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.Data.Entity.Utilities
+{
+    internal class IntegerSwitch : Switch
```
divega: @chrfin, @ajcvickers do you know how usable is System.Dianostics.Switch on .NET Core? From what I remember the last time I used one of these you configured in the app.config, but that isn't operational in .NET Core, so, are there other alternatives?

chrfin: Hmm, good point. I used a `Switch` as the class already uses `BooleanSwitch` (see `_applyTransformationsRegardlessOfSize`)...
I'll check it and if not working on .NET Core replace all (incl. the "old one") of them and update the pull-request.

chrfin: @ajcvickers @divega I checked it and Diego is right. `Switch` is useless in .NET Core as it is only working in-memory per instance.  
Now I'm not sure in which direction I should go:
1) Improve the `Switch`-implementation in CoreFX to be usable
2) Create a new `Switch`-implementation outside of CoreFX based on `Microsoft.Extensions.Configuration` (where could/should that be added? new package?)
3) Add `Microsoft.Extensions.Configuration` to EF6 and use the new config-system instead of the switches
4) Implement some global/static settings directly in EF6 changeable via code

What do you think?  
The best solution IMO would be using `Microsoft.Extensions.Configuration` in the CoreFX-`Switch`, but that is not possible AFAIK, or is it somehow?

divega: @ajcvickers or @bricelam may have good suggestions for this. Just some toughs: 

- We cannot add anything that depends on Microsoft.Extensions.Configuration, because the package targets .NET Standard 2.0, and doesn't support .NET Framework 4.0 and 4.5. Also, EF6's design doesn't define a good place to pass something like a configuration instance. 

- A change to System.Diagnostics.Switch that would require taking a hard dependency on System.Configuration.ConfigurationManager is unlikely to be accepted because System.Diagnostics.Switch is part of the .NET Core shared framework and this doesn't depend on ConfigurationManager by design. We also don't have a mechanism to do this with a soft dependency. 

- Elsewhere we are using AppContext.TryGetSwitch() in particular for _quirks_ you can use to opt out we usually add every time we change a behavior in a patch, but I think unfortunately that API doesn't exist prior to .NET Framework 4.6, so it wouldn't be an option for EF6. 

- All that said, EF6 already depends on System.Configuration.ConfigurationManager and we already have a few places in which we read flags from application settings:
https://github.com/aspnet/EntityFramework6/blob/8d2c24bc00aaa042a1e57eb1070f8ba23448e82a/src/EntityFramework/Core/Objects/ObjectContext.cs#L247
https://github.com/aspnet/EntityFramework6/blob/22a11d132306c257435fa7d6d75a31359cbbd05a/src/EFTools/EntityDesignerVersioningFacade/ReverseEngineerDb/SchemaDiscovery/EntityStoreSchemaGeneratorDatabaseSchemaLoader.cs#L59

I am not sure why the code in PlanCompiler uses BooleanSwitch. It is a decision that was made a long time ago when we didn't envision the possibility of running on .NET Core (or we didn't envision .NET Core for that matter :smile:), but also it sounds like a setting a user may want to change for non-diagnostic reasons.

chrfin:
@divega @ajcvickers @bricelam I now changed it to use `ConfigurationManager.AppSettings`.  
I also changed the existing switch - if that is a non-desired breaking-change please tell me and I revert that part...

src/EntityFramework/Core/Query/PlanCompiler/PlanCompiler.cs
```diff
@@ -41,23 +42,68 @@ internal class PlanCompiler
...
+        private static bool? _applyTransformationsRegardlessOfSize;
         // <summary>
-        // A boolean switch indicating whether we should apply transformation rules regardless of the size of the Iqt.
+        // A boolean value indicating whether we should apply transformation rules regardless of the size of the Iqt.
         // By default, the Enabled property of a boolean switch is set using the value specified in the configuration file.
         // Configuring the switch with a value of 0 sets the Enabled property to false; configuring the switch with a nonzero
         // value to set the Enabled property to true. If the BooleanSwitch constructor cannot find initial switch settings
         // in the configuration file, the Enabled property of the new switch is set to false by default.
         // </summary>
-        private static readonly BooleanSwitch _applyTransformationsRegardlessOfSize =
```
divega: I would recommend erring on the side of compatibility, so for anyone relying in the existing switch, configuration will still be applied. We can add the appSettings as a fallback.

src/EntityFramework/Core/Query/PlanCompiler/PlanCompiler.cs
```diff
@@ -41,23 +42,68 @@ internal class PlanCompiler
...
+
+                return _disableTransformationsRegardlessOfSize.Value;
+            }
+        }
 
+        private static int? _maxNodeCountForTransformations;
         // <summary>
-        // Determines the maximum size of the query in terms of Iqt nodes for which we attempt to do transformation rules.
-        // This number is ignored if applyTransformationsRegardlessOfSize is enabled.
+        // An integer switch specifying the maximum size of the query in terms of Iqt nodes for which we attempt to do transformation rules.
```
divega: Nit: integer value

src/EntityFramework/Core/Query/PlanCompiler/PlanCompiler.cs
chrfin: @divega I did restore the switch in parallel to the AppSettings

chrfin:
@divega May you can take another look? Thanks

chrfin:
@divega @ajcvickers @bricelam Ping after preview 9 ☺

divega:
@chrfin sorry for not responding earlier. We looked superficially at the PR and it is looking good, though we may need to do another pass to provide feedback. 

I just wanted to answer if this will make it into 6.3: at this time in a release (i.e. with no more previews in sight) we tend not to take changes of this nature as a heuristic to avoid introducing last regressions in the last minute. Based on this, I would like us to consider this for the next release.

chrfin:
Thanks for the feedback and I completely understand your reasoning :-)

divega:
@chrfin speaking of which, we will sadly have to revert the previous compilation perf improvement because of https://github.com/aspnet/EntityFramework6/issues/1217 :sad: We will of course consider it for the next release as well, once we have time to understand what caused it to break things.

chrfin:
@divega Sad to hear, but also understandable. Could you may keep the lowered `MaxNodeCountForTransformations` in, as this is already a huge improvement for my app (without that I could not use 6.3 final and would stick to the previews).
I will also take a look at #1217 to may get the changes back in at the first 6.4 preview...

divega:
cc @bricelam

ajcvickers:
Thanks for the contribution!

PR #1042, opened 2019-07-12
Move Enable-Migrations logic out-of-process

bricelam:
This is a back-port of the architecture used by EF Core. It enables F#, x64, .NET Core, and cross-targeting projects. PowerShell 3.0 is now required its for JSON support. This adds ef6.exe which will eventually replace migrate.exe

Part of #231


src/ef6/ResXResourceWriter.cs
```diff
@@ -0,0 +1,164 @@
...
+
+#if NETCOREAPP3_0
+
+using System.IO;
+using System.Text;
+
+namespace System.Resources
+{
+    // HACK: Work around dotnet/winforms#1342
+    internal class ResXResourceWriter : IDisposable
```
bricelam: lol, no worse than [MSBuild's version](https://github.com/microsoft/msbuild/blob/35ed18512aa6e05fde7e6d1daa487299349e8803/src/Tasks.UnitTests/ResourceHandling/ResXHelper.cs).

PR #1058, opened 2019-07-16
Make remaining Migrations commands out-of-process

### LINKED ISSUE #231: Make EF NuGet package / migrations and EF Tools work with new csproj format

bricelam:
Note, this removes migrate.exe; it has been superseded by ef6.exe

Resolves #231


src/EntityFramework/Migrations/Design/ToolingFacade.cs
```diff
@@ -27,7 +27,7 @@ namespace System.Data.Entity.Migrations.Design
     /// The App/Web.config file from the startup project is also copied
     /// to ensure that any configuration is applied.
     /// </summary>
-    // TODO: Move this functionality to System.Data.Entity.Infrastructure.Design.Executor
+    [Obsolete("Use System.Data.Entity.Infrastructure.Design.Executor instead.")]
```
bricelam: ❤

PR #1074, opened 2019-07-25
Update SDK to preview 7

JunTaoLuo:
Addresses aspnet/AspNetCore-Internal#2774

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "3.0.100-preview6-012264",
+    "dotnet": "3.0.100-preview7-012821",
     "runtimes": {
       "dotnet": [
         "$(MicrosoftNETCoreAppRuntimeVersion)"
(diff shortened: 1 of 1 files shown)


bricelam:
(Still wondering why we have to do this manually...)

dougbu:
@bricelam because the SDK doesn't exist in a Maestro++ channel and Arcade's minimum SDK lags, then lags some more.

bricelam:
Should we re-open dotnet/arcade#811? (or at least file a new issue)

dougbu:
The flow from Arcade works. The problem I was talking about was just them updating the SDK reactively i.e. when someone mentions it's lagging. Then, they need to notify Arcade users and wait a bit before changing. We generally want to move faster.

That said, I could open another Arcade PR to start their process. Is that important at this point?

bricelam:
I’m fine with it as it is (since I’m not the one doing it) it just feels like arcade didn’t deliver the feature we asked for

bricelam:
I.e. we should be able to opt into the latest SDK, not a “minimum required” version

PR #1095, opened 2019-07-29
Automatically include *.edmx files in SDK-style projects

### LINKED ISSUE #891: Include *.edmx files by default
Similar to how *.cs and *.resx files are [included by default](https://github.com/dotnet/sdk/blob/69d51538bdbdb03c758b924f887a6f596e590c7d/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#L26-L29) in SDK-style projects, we could include *.edmx files as `EntityDeploy` items.

bricelam:
Fixes #891

### DIFF
--- a/src/NuGet/EntityFramework/build/EntityFramework.props
+++ b/src/NuGet/EntityFramework/build/EntityFramework.props
@@ -1,0 +1,6 @@
+<Project>
+  <ItemGroup Condition="'$(EnableDefaultItems)' == 'True' And '$(EnableDefaultEntityDeployItems)' == 'True'">
+    <EntityDeploy Include="**/*.edmx" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <None Remove="**/*.edmx" Condition="'$(EnableDefaultNoneItems)' == 'True'" />
+  </ItemGroup>
+</Project>
--- a/src/NuGet/EntityFramework/build/EntityFramework.targets
+++ b/src/NuGet/EntityFramework/build/EntityFramework.targets
@@ -1,5 +1,9 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <PropertyGroup>
+    <EnableDefaultEntityDeployItems Condition="'$(EnableDefaultEntityDeployItems)' == ''">True</EnableDefaultEntityDeployItems>
+  </PropertyGroup>
+
   <!-- This .targets file can be used by updating Microsoft.Common.targets to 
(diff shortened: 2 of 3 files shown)


bricelam:
@pranavkm @javiercn Any gotchas?

javiercn:
I’d recommend having buildTransitive pros to point to buildMultitargeting prod and that to build props.

I think you added the elements correctly AFAIK I don’t remember if there were 2 flags or 1 for content, as 1 for default items and a more specific one for web content.

WRT msbuild im not sure if you need to compare against 'True' or 'true'. I always do the latter, but not sure if it matters.

CZEMacLeod:
It would be useful to have an EF6 specific property to turn this behavior off. I use CodeFirst and Migrations, but also extract the target EDMX from the migration resource file and set it to be parented to the migration. This means that I can easily look at the generated model visually and check if it is correct or use it for reference, print it, share it with my team etc.
With this change (and I think something in VS2019 when it adds EDMX files now even with EF6.2) I have to effectively override this and set all my EDMX 'reference' files back to None.
Maybe we could use the standard Migrations or Migrations.* folders as an exclusion?

bricelam:
@javiercn AFAICT, buildMultitargeting is only needed to affect restore and pack. These are effectively embedded resources so they only matter when compiling for a specific framework.

MSBuild is case insensitive. I like to channel my inner VB developer when writing conditions since it's `And` not `&&`.

I'll add an off switch for edmx files similar to the existing ones. This should enable @CZEMacLeod's scenario

javiercn:
> @javiercn AFAICT, buildMultitargeting is only needed to affect restore and pack. These are effectively embedded resources so they only matter when compiling for a specific framework.

That's fair. I normally do the chaining transitive -> multitargeting -> build because its safe in all situations.

> MSBuild is case insensitive. I like to channel my inner VB developer when writing conditions since it's `And` not `&&`

It's fair

> I'll add an off switch for edmx files similar to the existing ones. This should enable @CZEMacLeod's scenario

This is a good thing to have. The razor SDK also does it.

bricelam:
Oh I guess the other place where you need multitargeting is when you have an entry-point target. Otherwise `msbuild /t:MyTarget` wouldn't work on multi-targeting projects.

bricelam:
Updated. @CZEMacLeod add this to your project to prevent it from hapenning by default (you'll still be able to manually change them to EntityDeploy items if you want)
``` xml
<PropertyGroup>
  <EnableDefaultEntityDeployItems>False</EnableDefaultEntityDeployItems>
</PropertyGroup>
```

PR #1122, opened 2019-08-06
Bug 898645. Workaround NRE from runtime.

lajones:
Workaround for an issue a couple of users are seeing where we don't guard against null being returned and end up with an NRE.


lajones:
Correction. Should be fix for bug 951609 (not 898645 - although that may be related).

PR #1124, opened 2019-08-07
Add paragraphs to make remarks more readable

ite-klass:
Without the `para` tags the text will be collapsed into a single, huge
paragraph making it very hard to read with how much content is in there.

This change adds the `para` tags to declare paragraphs where they were
seemingly intended anyway, except for one instance where a new
paragraph/split is introduced.

https://docs.microsoft.com/en-us/dotnet/api/system.data.entity.dbcontext?view=entity-framework-6.2.0#remarks


ajcvickers:
Thanks for the contribution!

PR #1125, opened 2019-08-07
Fixes System.ArgumentException when GetAncestor returns the root node

iosebyte:
The method 'GetAncestor' throws an ArgumentException when the found ancestor is the root node, because hierarchyStr is '//', which is not a valid value.

Now this method checks if the level of the current node and the n parameter is the same, and in this case returns the root node.
Added new test case to check this behaviour


ajcvickers:
Thanks!

PR #1153, opened 2019-08-16
Allow creation of a DbMigrator with an explicit DbContext instance

### LINKED ISSUE #522: DbMigrator should avoid connecting to master and should use existing connection if possible
The current implementation causes issues, e.g. when using ADD authentication with Azure SQL DB.

From @ajcvickers's email:

> ... it seems that there is an issue with the database existence check that is done by the DbMigrator—it is attempting to connect to the master database, which fails.  We had updated other existence checks to not require this, but it looks like DbMigrator was either missed or there is some reason it couldn’t be done here... 

Also, there is apparently a code path there that may leave a connection open.

ajcvickers:
This prevents migrations having to attempt to create connections from connection strings.

Fixes #522

### DIFF
--- a/src/EntityFramework/Migrations/DbMigrator.cs
+++ b/src/EntityFramework/Migrations/DbMigrator.cs
@@ -94,10 +94,17 @@ namespace System.Data.Entity.Migrations
             Check.NotNull(configuration.ContextType, "configuration.ContextType");
         }
 
+        /// <summary>
+        /// Initializes a new instance of the DbMigrator class using the supplied context.
+        /// Use this constructor when applying migrations from code to avoid having migrations attempt
+        /// to create a context/connection for you.
+        /// </summary>
+        /// <param name="configuration"> Configuration to be used for the migration process. </param>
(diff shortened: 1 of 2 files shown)


CZEMacLeod:
Really useful change - I can finally get rid of the following code from my MigrationDBInitializer class... 😃 
```
Private Shared ReadOnly _createNewMigrator As Reflection.ConstructorInfo
Shared Sub New()
    _createNewMigrator = GetType(Entity.Migrations.DbMigrator).GetConstructor(Reflection.BindingFlags.NonPublic Or Reflection.BindingFlags.Instance,
                                                                            Nothing,
                                                                            {GetType(Entity.Migrations.DbMigrationsConfiguration),
                                                                            GetType(DbContext)},
                                                                            Nothing)
End Sub
Protected Shared Function GetNewMigrator(ByVal configuration As Entity.Migrations.DbMigrationsConfiguration,
                                            ByVal usersContext As DbContext) As Entity.Migrations.DbMigrator
    Return _createNewMigrator.Invoke({configuration, usersContext})
End Function
```

PR #1162, opened 2019-08-19
Fix PMC error when no config file

### LINKED ISSUE #1143: Enable migrations - preview 8
VS 16.3.0 preview 2
.NET Core 3 preview 8
.NET 4.8

Create 4 class libraries:

ClassLibrary.NET (targets .NET 4.7.2)
ClassLibraryNetCore (targets .NET Core 3 p8)
ClassLibraryStandard(.NET standard 2.1 p8)
ClassLibraryMulti (multi target all)

Add EF 6.3 p8 to all of them and add an empty DbContext.

Try to use enable-migrations on each of them:

> enable-migrations -project classlibrary.net -startupproject classlibrary.net

 fails with:

image

> enable-migrations -project classlibrary.net -startupproject **classlibrarynetcore**

oddly this works and enables migrations on the .net project.

> enable-migrations -project classlibrarynetcore -startupproject classlibrarynetcore

works.

> enable-migrations -project classlibrarynetstandard -startupproject classlibrarynetstandard

fails with:

image

Which seems odd since surely the .NET core tools can handle a .NET standard assembly?

> enable-migrations -project classlibrarymulti -startupproject classlibrarymulti

works.

Should the tools just work in all of these scenarios?

Demo solution: [EfMigrate.zip](https://github.com/aspnet/EntityFramework6/files/3502732/EfMigrate.zip)

bricelam:
Fixes #1143

### DIFF
--- a/src/NuGet/EntityFramework/tools/EntityFramework6.psm1
+++ b/src/NuGet/EntityFramework/tools/EntityFramework6.psm1
@@ -47,7 +47,7 @@ function Add-EFProvider
         [parameter(Position = 2, Mandatory = $true)]
         [string] $TypeName)
 
-    $configPath = GetConfigPath($Project)
+    $configPath = GetConfigPath $Project
     if (!$configPath)
     {
         return
@@ -110,7 +110,7 @@ function Add-EFDefaultConnectionFactory
         [string] $TypeName,
         [string[]] $ConstructorArguments)
 
-    $configPath = GetConfigPath($Project)
+    $configPath = GetConfigPath $Project
     if (!$configPath)
     {
         return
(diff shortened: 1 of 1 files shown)


src/NuGet/EntityFramework/tools/EntityFramework6.psm1
```diff
@@ -1205,7 +1205,11 @@ function GetConfigPath($project)
         $configFileName = 'app.config'
     }
 
-    return GetProperty $project.ProjectItems.Item($configFileName).Properties 'FullPath'
```
bricelam: Note, new csproj returned null if the item didn't exist, but old csproj threw. Missed this since the .NET Framework Console template includes an app.config

src/NuGet/EntityFramework/tools/EntityFramework6.psm1
```diff
@@ -47,7 +47,7 @@ function Add-EFProvider
         [parameter(Position = 2, Mandatory = $true)]
         [string] $TypeName)
 
-    $configPath = GetConfigPath($Project)
+    $configPath = GetConfigPath $Project
```
bricelam: These weren't wrong, just confusing. A second argument would have to be added after the closing parameter `F($x) $y $z`

PR #1274, opened 2019-09-18
Alternative cleanup databases script

AndriySvyryd:


azure-pipelines.yml
```diff
@@ -55,25 +55,15 @@ stages:
                                       /p:DotNetPublishUsingPipelines=$(_PublishUsingPipelines)
                                       /p:DotNetArtifactsCategory=$(_DotNetArtifactsCategory)
             steps:
+              - checkout: self
+                clean: true
               - powershell: |
                   SqlLocalDB stop MSSQLLocalDB -i
+                  SqlLocalDB delete MSSQLLocalDB
+                  Remove-Item "$env:USERPROFILE\*.mdf"
+                  Remove-Item "$env:USERPROFILE\*.ldf"
```
bricelam: Are they all here? We used to put Northwind files in a different location (next to master)

AndriySvyryd: We'll see once this runs

PR #1276, opened 2019-09-19
Update to 5.0 SDK

wtgodbe:
Part of https://github.com/aspnet/AspNetCore-Internal/issues/3016

CC @ajcvickers @JunTaoLuo @dougbu

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -11,12 +11,12 @@
     <KnownAppHostPack 
       Include="@(KnownAppHostPack->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
       TargetFramework="$(DefaultNetCoreTargetFramework)"
-      Condition="!(@(KnownAppHostPack->AnyHaveMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)')))">
+      Condition="!(@(KnownAppHostPack->AnyHaveMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)')))"
       />
     <KnownFrameworkReference
       Include="@(KnownFrameworkReference->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
       TargetFramework="$(DefaultNetCoreTargetFramework)"
-      Condition="!(@(KnownFrameworkReference->AnyHaveMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)')))">
+      Condition="!(@(KnownFrameworkReference->AnyHaveMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)')))"
       />
   </ItemGroup>
 
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- Workaround while there is no 5.0 SDK available, copy known apphost/framework reference info from 3.0 -->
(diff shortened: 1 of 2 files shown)


Directory.Build.targets
```diff
@@ -6,18 +6,6 @@
...
-    <!-- Workaround while there is no 5.0 SDK available, copy known apphost/framework reference info from 3.0 -->
-    <KnownAppHostPack 
-      Include="@(KnownAppHostPack->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
-      TargetFramework="netcoreapp5.0"
-      />
-    <KnownFrameworkReference
-      Include="@(KnownFrameworkReference->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
-      TargetFramework="netcoreapp5.0"
-      />
-  </ItemGroup>
```
dougbu: Instead make this logic conditional on lack of `%(TargetFramework) == $(DefaultNetCoreTargetFramework)` items

Directory.Build.targets
```diff
@@ -6,18 +6,6 @@
...
-
   <ItemGroup>
     <!-- Reference base shared framework at incoming dependency flow version, not bundled sdk version. -->
     <FrameworkReference
```
dougbu: Suggest instead updating `@(KnownFrameworkReference)` items

Directory.Build.targets
```diff
@@ -10,17 +10,19 @@
     <!-- Workaround while there is no 5.0 SDK available, copy known apphost/framework reference info from 3.0 -->
     <KnownAppHostPack 
       Include="@(KnownAppHostPack->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
```
wtgodbe: I couldn't figure out a good way to generalize this to "Include KnownAppHostPack items where TargetFramework != DefaultNetCoreTargetFramework" - as far as I can tell, WithMetadataValue is inclusive only: https://docs.microsoft.com/en-us/visualstudio/msbuild/item-functions?view=vs-2019

dougbu: Combine `WithMetadataValue` and `Count` item functions

wtgodbe: I'm not sure that works either - I can include items based on `WithMetadataValue`, but I could only use `Count` in a condition. I could include all items in `KnownFrameworkReference` here, then condition on: `KnownFrameworkReference` has no Items with MetadataValue `TargetFramework` = `DefaultNetCoreTargetFramework`, AND `TargetFramework` starts with `Netcoreapp`, but then if we used a 5.0 SDK in vNext (6.0?), we'd be adding 6.0 entries for both the 3.0 & 5.0 netcoreapp references in the 5.0 SDK, which seems wrong.

dougbu:
```
Condition=" @(KnownAppHostPack->WithMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)->Count() == 0 "
```
&hellip;
```
AND %(KnownAppHost.TargetFramework) == 'netcoreapp3.0'
```
and use `$(DefaultNetCoreTargetFramework)` in the added item.

wtgodbe: In that case, I think this is already in the desired state - We add an entry that duplicates the `netcoreapp3.0` entry, but with `TargetFramework = $(DefaultNetCoreTargetFramework)`, only when the ItemGroup doesn't already have an item with `TargetFramework = $(DefaultNetCoreTargetFramework)`.

wtgodbe:
@dougbu @JunTaoLuo does this look good to go?

PR #1286, opened 2019-09-23
Switch to embedded PackageIcon

wtgodbe:
Part of https://github.com/aspnet/AspNetCore-Internal/issues/3146

Local build passes the smell test with these changes. I'm still unsure on whether or not we need to include a reference to the package icon in the `.nuspec` - the "old way" did so automatically, the "new way" does not (https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file). CC @tmat who may know more about that question.


tmat:
@nkolev92

nkolev92:
That's really up to you at this point. 

NuGet.org will eventually transition to a world where it only accepts packages with a self contained icon, but other feeds may not implement this soon if ever.

dougbu:
> only accepts packages with a self contained icon,

I'm not sure this answers the question we had. **For NuGet.org**, should the .nuspec reference the icon file?

nkolev92:
The newly icon metadata has to be used for NuGet.org.

wtgodbe:
> The newly icon metadata has to be used for NuGet.org.

@nkolev92 what do you mean by this? I followed the example from https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file in this PR, which resulted in an icon file in the root of the .nupkg, but no reference to it in the `.nuspec`. Is that sufficient?

nkolev92:
The adding of the propery should burn it into the nuspec. 

```xml
    <PackageIcon>icon.png</PackageIcon>
```

If that's not the case, make sure a high enough version of the SDK is being sued.
I think this was added in 3.0.0-preview8

tmat:
@nkolev92 That's assuming the nuspec is auto-generated, correct? It seems like the projects in this repo have hand written nuspecs. @wtgodbe Perhaps it would be good time to migrate to auto-generated nuspecs?

wtgodbe:
Ah, looks like you're correct. I made this change in several repos, all of which auto-generate nuspec's by default, except for a few exceptions (such as the manual ones in this PR). The ones which were auto-generated do have the `icon` metadata in their nuspec. I'll add an `icon` element to the manual `.nuspec`'s, then open an issue to migrate all `.nuspec`'s to be auto-generated.

wtgodbe:
With the last change, this now includes the icon metadata in all .nuspec's, including the manually generated ones - @dougbu PTAL

wtgodbe:
Closing in favor of https://github.com/aspnet/EntityFramework6/pull/1291

PR #1291, opened 2019-09-24
[release/6.3] Switch to embedded package icon

wtgodbe:
Contributes to aspnet/AspNetCore-Internal#3103

PR #1312, opened 2019-09-27
Disable DependentUpon convention for EntityFramework.csproj

rainersigwald:
Code generated by Resources.tt->Resources.cs assumes a resource naming convention that isn't compatible with the DependentUpon convention, so disable it for this project.


wtgodbe:
Still seeing the test failures:

> System.Resources.MissingManifestResourceException : Could not find the resource \"System.Data.Entity.Properties.Resources.resources\" among the resources \"System.Data.Entity.Resources.Strings.resources\", \"System.Data.Resources.AnnotationSchema.xsd\", \"System.Data.Resources.CodeGenerationSchema.xsd\", \"System.Data.Resources.CSDLSchema_1.xsd\", \"System.Data.Resources.CSDLSchema_1_1.xsd\", \"System.Data.Resources.CSDLSchema_2.xsd\", \"System.Data.Resources.CSDLSchema_3.xsd\", \"System.Data.Resources.EntityStoreSchemaGenerator.xsd\", \"System.Data.Resources.SSDLSchema.xsd\", \"System.Data.Resources.SSDLSchema_2.xsd\", ... embedded in the assembly \"EntityFramework\", nor among the resources in any satellite assemblies for the specified culture. Perhaps the resources were embedded with an incorrect name.

src/EntityFramework/EntityFramework.csproj
```diff
@@ -7,6 +7,9 @@
     <Description>EntityFramework.dll</Description>
     <AssemblyVersion>6.0.0.0</AssemblyVersion>
     <DefineConstants>$(DefineConstants);INTERNALS_INVISIBLE</DefineConstants>
+
+    <!-- Properties\Resources.tt expects a path that does not match the class in Properties\Resources.cs -->
```
nguerrera: Is it worth changing the code to use the conventional name?

rainersigwald: This is a great question I leave up to the maintainers. It didn't seem trivial, but I don't know the history of the `Resources.tt` approach or the details of its implementation.

rainersigwald:
So it looks like I fixed _almost_ all the errors, but there's two instances of this:

```
Actual string does not match the message defined in resources.
Expected:'Spatial readers can only be produced from readers of type SqlDataReader. A reader of type Castle.Proxies.DbDataReaderProxy was provided.',
Actual:'The provider did not return a 'DbSpatialServices' instance. In order to use the 'DbGeography' or 'DbGeometry' spatial types the EF provider being used must support spatial types and all prerequisites for the provider must be installed. See http://go.microsoft.com/fwlink/?LinkId=287183 for details.'
Expected: True
Actual: False
```

@wtgodbe do you think that's resource-related?

rainersigwald:
@ajcvickers, do you know (or know who might know) what might be up with that remaining error? Is it likely to be related to resource naming?

ajcvickers:
@bricelam @rainersigwald @dougbu This was happening while I was on vacation; sorry for missing it. Does this change need to go in for the 3.1 release train?

dougbu:
I believe @wtgodbe incorporated this proposal in #1291. @rainersigwald / @wtgodbe do we need this PR for something else?

ajcvickers:

PR #1338, opened 2019-10-03
Partial fix for build problems.

lajones:
Some updates to fix the build:

1. updates so the paths build on the local machine as well as on CI
2. update to 16.3 version of `Microsoft.VisualStudio.Threading`
3. ignore CA1305 - this is a new CodeAnalysis rule to which we have many exceptions. It's about if you convert an int (or long etc) to string automatically in e.g. `Debug.Assert(someCount < 2, "Actual count is " + someCount)`, they want you to provide an `IFormatProvider` every time. We have many of these - and the exact format of the string never matters - the number we're putting out would be in the single digits anyway and we don't really care about the exact format of the string even if it got into the 1000's. Simplest answer is just to disable this error.

Note: still having problems with the build after that but would like to get this much in before I go on vacation.

### DIFF
--- a/.nuget/NuGet.targets
+++ b/.nuget/NuGet.targets
@@ -27,13 +27,13 @@
     
     <PropertyGroup Condition=" '$(OS)' == 'Windows_NT'">
         <!-- Windows specific commands -->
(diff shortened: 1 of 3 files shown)


maumar:
:shipit:

Issue #1378, opened 2019-10-13
Potential cross-platform path handling issue in EntityDeploy MSBuild task


mistachkin:
The OutputCMS method in the source file "src/Microsoft.Data.Entity.Build.Tasks/EntityDeploy.cs" has the following code:

**FileInfo outputFile = new FileInfo(Path.Combine(topLevelOutputDir.FullName, inputFileRelativePath));**

And then a bit later in the same method:

**string outputDirPath = outputDir.FullName + "\\\\";**

I think hard-coding the backslash here is causing the problem I'm seeing with my Linux build:

CSC : error CS1566: Error reading resource 'NorthwindModel.EF6.NetStandard21.csdl' -- 'Could not find file './testlinq/obj/Debug/netcoreapp3.0/NorthwindModel.EF6.NetStandard21.csdl'.' [./testlinq/testef6.NetStandard21.csproj]

Within the "**obj/Debug/netcoreapp3.0**" directory, I see the following files and directories:

(directory) edmxResourcesToEmbed
(file) **edmxResourcesToEmbed\NorthwindModel.EF6.NetStandard21.ssdl**
(file) **edmxResourcesToEmbed\NorthwindModel.EF6.NetStandard21.msl**
(file) **edmxResourcesToEmbed\NorthwindModel.EF6.NetStandard21.csdl**
(file) testef6.NetStandard21.assets.cache
(file) testef6.NetStandard21.csproj.FileListAbsolute.txt
(file) testef6.NetStandard21.csprojAssemblyReference.cache

Notice the three files that have a backslash within their name?  That is the root of the problem, which I'm pretty sure was caused by the errant code in the MSBuild task.

I think the solution may be as simple as using the System.IO.Path.DirectorySeparatorChar field instead of hard-coding the backslash; alternatively, the Path.Combine method could potentially be used (again) in this method to avoid using the character at all.

FYI: This issue is currently **blocking** the release of .NET Core 3.0 support for System.Data.SQLite.

ajcvickers:
@bricelam Thoughts?

mistachkin:
Update: I think part of the problem here is actually a bad interaction with the private **CreateTaskItem** method within the same file.  It searches for **Path.DirectorySeparatorChar**, not backslash.  Seemingly, this would lead to an incorrect **LogicalName** being set on the created **TaskItem**.

mistachkin:
I think pull request #1385 should contain the necessary changes to fix this issue.

[...]

PR #1385, opened 2019-10-16
Always use Path.Combine when building fully qualified file names.  Fix for issue #1378.

mistachkin:
These changes remove the hard-coding of the backslash as a directory separator in the EntityDeploy MSBuild task.  This is necessary for portability to non-Windows operating systems.  See issue #1378.


bricelam:
Joe! lol, I just realized this was you. Long time no see! Thanks for the contribution

PR #1446, opened 2019-11-05
Opt-out of dependency flow

bricelam:
I intend to update to the latest 3.1 versions as they're released to NuGet.org.

This should keep `master` in good working order as we decouple this repo from the 5.0 release.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,32 +1,3 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
-
-  <!-- Workaround while there is no 5.0 SDK available, suppress unsupported version error -->
-  <PropertyGroup>
-    <NETCoreAppMaximumVersion>5.0</NETCoreAppMaximumVersion>
-  </PropertyGroup>
-
-  <ItemGroup>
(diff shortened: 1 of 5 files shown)


Directory.Build.targets
```diff
@@ -1,32 +1,3 @@
...
-      Include="@(KnownAppHostPack->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
-      TargetFramework="$(DefaultNetCoreTargetFramework)"
-      Condition="!(@(KnownAppHostPack->AnyHaveMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)')))"
-      />
-    <KnownFrameworkReference
-      Include="@(KnownFrameworkReference->WithMetadataValue('TargetFramework', 'netcoreapp3.0'))"
-      TargetFramework="$(DefaultNetCoreTargetFramework)"
-      Condition="!(@(KnownFrameworkReference->AnyHaveMetadataValue('TargetFramework', '$(DefaultNetCoreTargetFramework)')))"
-      />
-  </ItemGroup>
```
dougbu: This block will no-op when using any recent 3.1.* or greater .NET Core SDK. Suggest keeping it in reserve unless everyone is absolutely sure this repo will never need a newer TFM than the chosen SDK supports intrinsically.

Issue #557, opened 2018-06-15
Memory leak in InitializerMetadata


SocSoftSpb:
Some inheritors of System.Data.Entity.Core.Objects.ELinq.InitializerMetadata (ProjectionInitializerMetadata and ProjectionNewMetadata) causes Memory leak. Instances of ProjectionInitializerMetadata and ProjectionNewMetadata can hold references to some closures and other objects via ConstantExpression's in subtree of NewExpression or MemberInitExpression.

Screenshot from dotMemory with retention graph example is attached.
![leak](https://user-images.githubusercontent.com/34467689/41454426-06ec8794-7082-11e8-8434-85ea1710b4c3.PNG)

These types of InitializerMetadata does not requires of full expression tree, only "shapes" of NewExpression or MemberInitExpression is used in code.

Proposed solution:
In constructors of ProjectionInitializerMetadata and ProjectionNewMetadata create reduced copy of original NewExpression or MemberInitExpression, with replace unused subtrees to "dummy" constant expressions with matched types.

ajcvickers:
This issue has been closed because EF6 is no longer being actively developed. We are instead focusing on stability of the codebase, which means we will only make changes to address security issues. See the repo [README](https://github.com/dotnet/ef6/blob/main/README.md) for more information.

PR #1542, opened 2020-01-06
Fix: Memory leak in InitializerMetadata #557

SocSoftSpb:
Fix of memory leak & test reproducing memory leak with WeakReference.


ajcvickers:

PR #1595, opened 2020-02-18
Fix messages erroneously mention EF Core

leotsarev:


leotsarev:
Guys, I understand this is not very important PR, and EF6 is essentially in maintenance mode-only, but "maintenance mode" is not same as "not maintained".

ajcvickers:

PR #1610, opened 2020-03-08
[release/6.3] Update hosted and BYOC agents used in our builds

dougbu:
- dotnet/aspnetcore-internal#3540

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,10 +36,10 @@ stages:
             pool:
               ${{ if eq(variables['System.TeamProject'], 'public') }}:
                 name: NetCorePublic-Pool
-                queue: BuildPool.Windows.10.Amd64.VS2017.Open
+                queue: BuildPool.Server.Amd64.VS2017.Open
               ${{ if ne(variables['System.TeamProject'], 'public') }}:
                 name: NetCoreInternal-Pool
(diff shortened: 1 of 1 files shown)


dougbu:
**NO-MERGE** because the branches aren't open

azure-pipelines.yml
```diff
@@ -13,7 +13,7 @@ variables:
 trigger:
   - master
   - release/*
-  - internal/release/*
+  - internal/release/6.*
```
Pilchie: Why this change?

dougbu: I added an extra commit in an attempt to figure out why the database cleanup was failing in the release branches but succeeded (I thought) in 'master'. Since the 'master' validation build now failed, I suspect something else is going on -- perhaps LocalDb is slightly different or missing on these agents.

dougbu: Will remove that commit once I understand what's going on w/ the failing database cleanups.

dougbu: The commit adding the unrelated changes is gone now. Much less to see here 😈

dougbu:
Though removing the extra commit triggered another build, I doubt that will succeed. Suspect LocalDB issues in the Server agents -- as discussed in #1612.

dougbu:
No need to update this branch

PR #1631, opened 2020-03-27
Detach databases in cleanup script

AndriySvyryd:


AndriySvyryd:
@dougbu `sqlcmd` isn't in the path

azure-pipelines.yml
```diff
@@ -69,14 +69,8 @@ stages:
...
                 clean: true
-              - powershell: |
-                  SqlLocalDB stop
-                  SqlLocalDB delete
-                  Remove-Item "$env:USERPROFILE\*.mdf"
-                  Remove-Item "$env:USERPROFILE\*.ldf"
-                  SqlLocalDB create
-                  SqlLocalDB start
-                displayName: Cleanup databases
+              - script: tools\CleanMSSQLLocalDB.cmd
```
dougbu:
```suggestion
              - powershell: Get-ChildItem -Recurse -ErrorAction SilentlyContinue 'C:\Program Files*\sqlcmd.exe'
              - script: tools\CleanMSSQLLocalDB.cmd
```
Let's see if `sqlcmd.exe` is installed at all. If it is, can add directory to the path. If not, we can ask for that addition to the agent configuration.

To confirm: EFCore builds don't use `sqlcmd.exe` on the CI?

azure-pipelines.yml
```diff
@@ -69,14 +69,9 @@ stages:
...
                 clean: true
-              - powershell: |
-                  SqlLocalDB stop
-                  SqlLocalDB delete
-                  Remove-Item "$env:USERPROFILE\*.mdf"
-                  Remove-Item "$env:USERPROFILE\*.ldf"
-                  SqlLocalDB create
-                  SqlLocalDB start
-                displayName: Cleanup databases
+              - powershell: Get-ChildItem -Recurse -ErrorAction SilentlyContinue 'C:\Program Files*\sqlcmd.exe'
```
dougbu:
```suggestion
              - powershell: Get-ChildItem "C:\Program Files\Microsoft SQL Server\*\Tools\Binn\sqlcmd.exe",
                  "C:\Program Files\Microsoft SQL Server\Client SDK\ODBC\*\Tools\Binn\sqlcmd.exe" |
                Sort-Object -Property @{Expression={$_.FullName.Length}; Descending=$true},
                  @{Expression={$_.FullName}; Descending=$true} |
                Select-Object -First 1 -Property FullName |
                Foreach-Object {"##vso[task.prependpath]$_"}
```
This should be fairly resilient if new SQL Server versions are added though it may be wrong to prefer Client SDK over all regular SQL Server installations.

azure-pipelines.yml
```diff
@@ -69,14 +69,15 @@ stages:
...
                 clean: true
-              - powershell: |
-                  SqlLocalDB stop
-                  SqlLocalDB delete
-                  Remove-Item "$env:USERPROFILE\*.mdf"
-                  Remove-Item "$env:USERPROFILE\*.ldf"
-                  SqlLocalDB create
-                  SqlLocalDB start
-                displayName: Cleanup databases
+              - powershell:>
```
dougbu:
```suggestion
              - powershell: >
```

dougbu:
So, either my addition or yours was enough. Suggest sticking a `$env:Path -split ';'` Powershell command in there somewhere to figure it out. If it's yours, lemme know and I'll get back to YAML hacking (the PowerShell portion works fine locally).

tools/CleanMSSQLLocalDB.cmd
```diff
@@ -0,0 +1,6 @@
+echo %PATH%
```
dougbu: Not sure I've seen the current error before. Would restoring `@echo off` get the script working again?
```suggestion
@echo off
```

AndriySvyryd: I am not sure how `echo` would produce that error, perhaps %PATH% contains newline characters

dougbu: I don't know either. I'll see if there's a way to make sure prepended paths are syntactically correct later tonight. The dog is getting insistent about her walk&hellip;

Separately, should we retarget this to 'release/6.4'?

AndriySvyryd: Yeah, it wasn't coming from that line. Anyway seems like it's working 🤷‍♂

dougbu: (Have to admit it's been fun running up the number of commits in this "simple" PR 😈)

AndriySvyryd: Now you understand why we were hesitant to do this at all

azure-pipelines.yml
```diff
@@ -69,14 +69,13 @@ stages:
...
-                  Remove-Item "$env:USERPROFILE\*.mdf"
-                  Remove-Item "$env:USERPROFILE\*.ldf"
-                  SqlLocalDB create
-                  SqlLocalDB start
-                displayName: Cleanup databases
+              - powershell: >
+                  Get-ChildItem -ErrorAction SilentlyContinue -Recurse 'C:\Program Files\Microsoft SQL Server\sqlcmd.exe' |
+                  Sort-Object -Property @{Expression={$_.FullName.Length}; Descending=$true},@{Expression={$_.FullName}; Descending=$true} |
+                  Select-Object -First 1 |
+                  Foreach-Object {"##vso[task.prependpath]$($_.DirectoryName)"}
```
dougbu: Still hitting the error about Microsoft. Is it as simple as&hellip;?
```suggestion
                  Foreach-Object {Write-Host "##vso[task.prependpath]$($_.DirectoryName)"}
```

AndriySvyryd: I don't think it's related as it shows up even without the powershell script

azure-pipelines.yml
```diff
@@ -69,14 +69,13 @@ stages:
...
-                  Remove-Item "$env:USERPROFILE\*.ldf"
-                  SqlLocalDB create
-                  SqlLocalDB start
-                displayName: Cleanup databases
+              - powershell: >
+                  Get-ChildItem -ErrorAction SilentlyContinue -Recurse 'C:\Program Files\Microsoft SQL Server\sqlcmd.exe' |
+                  Sort-Object -Property @{Expression={$_.FullName.Length}; Descending=$true},@{Expression={$_.FullName}; Descending=$true} |
+                  Select-Object -First 1 |
+                  Foreach-Object {Write-Host "##vso[task.prependpath]$($_.DirectoryName)"}
+              - script: tools\CleanMSSQLLocalDB.cmd
```
dougbu:
```suggestion
              - powershell: $env:Path -split ';'
              - script: tools\CleanMSSQLLocalDB.cmd
```

azure-pipelines.yml
```diff
@@ -69,14 +69,14 @@ stages:
...
-                  Remove-Item "$env:USERPROFILE\*.ldf"
-                  SqlLocalDB create
-                  SqlLocalDB start
-                displayName: Cleanup databases
+              - powershell: >
+                  Get-ChildItem -ErrorAction SilentlyContinue -Recurse 'C:\Program Files\Microsoft SQL Server\sqlcmd.exe' |
+                  Sort-Object -Property @{Expression={$_.FullName.Length}; Descending=$true},@{Expression={$_.FullName}; Descending=$true} |
+                  Select-Object -First 1 |
+                  Foreach-Object {Write-Host "##vso[task.prependpath]$($_.DirectoryName)"}
+              - powershell: $env:Path -split ';'
```
dougbu: Revert and give previous step a nice name
```suggestion
                displayName: Add SqlCmd to path
```

tools/CleanMSSQLLocalDB.cmd
```diff
@@ -0,0 +1,6 @@
+@echo off
+sqlcmd -S "(localdb)\mssqllocaldb" -i "%~dp0DropAllDatabases.sql" -o "DropAll.sql"
+sqlcmd -S "(localdb)\mssqllocaldb" -i "DropAll.sql"
```
dougbu:
```suggestion
type DropAll.sql
sqlcmd -S "(localdb)\mssqllocaldb" -i "DropAll.sql"
```

dougbu:
:ok: @AndriySvyryd the real problem is finally exposed at https://dev.azure.com/dnceng/public/_build/results?buildId=577356&view=logs&j=0bc77094-9fcd-5c38-f6e4-27d2ae131589&t=e5c9247b-367b-556a-ffed-53eb5f68dbb5&l=14
```
Sqlcmd: Error: Microsoft SQL Server Native Client 11.0 : Unable to complete login process due to delay in opening server connection.
```
This is what's written to the DropAll.sql file. Does this mean the commands should include `-l 32` or so (4x the default)? Or, should the script check whether `"%LOCALAPPDATA%\Microsoft\Microsoft SQL Server Local DB\Instances\MSSQLLocalDb" exists before it does anything? (Previous runs on the agent may not have created the default instance.)

----
Separately, should the command be invoked w/ `-r1 >DropAll.sql` instead of the `-o` option to avoid writing errors and other messages into the next script?

AndriySvyryd:
@dougbu Feel free to add more time, check for MSSQLLocalDb (or call `SqlLocalDB start`) and pipe out the errors

tools/CleanMSSQLLocalDB.cmd
```diff
@@ -0,0 +1,7 @@
+@echo off
+sqlcmd -S "(localdb)\mssqllocaldb" -i "%~dp0DropAllDatabases.sql" -o "DropAll.sql"
```
dougbu:
```suggestion
SqlLocalDB start
sqlcmd -S "(localdb)\mssqllocaldb" -i "%~dp0DropAllDatabases.sql" -l 32 -r0 >"DropAll.sql"
echo Will execute:
```

tools/CleanMSSQLLocalDB.cmd
```diff
@@ -0,0 +1,9 @@
+@echo off
+SqlLocalDB start
+sqlcmd -S "(localdb)\mssqllocaldb" -i "%~dp0DropAllDatabases.sql" -l 32 -r0 >"DropAll.sql"
+echo Will execute:
+type DropAll.sql
```
dougbu: Up to you whether we need these two lines or, if we need them, whether to add something separating the script content from whatever follows.

AndriySvyryd: I doubt anyone will see this output unless something goes wrong

AndriySvyryd:
I don't have permissions to merge this

dougbu:
> I don't have permissions to merge this

:ok: I'll get this in after the CI does its thing.

dougbu:
Single test failure but no reason to rush things. Reran the job.

dougbu:
Many thanks @AndriySvyryd

PR #1650, opened 2020-04-18
Issue #561: Cast error, not enough information (add column info)

DenisBalan:
Calling stored procedures from entity framework involves type mapping.
Very often procedure result schema and code-behind class are not in sync.
Debugging this problem is a mess, we even had to include EF source code to understand problematic columns.
Error example:
```
The specified cast from a materialized 'System.Int32' type to the 'System.Double' type is not valid.
```
This PR aims to add more info to such errors so the exception would contain:
```
The specified cast from a materialized column 'column0' with type 'System.Int32' to the 'System.Double' type is not valid.
```
Notice column name present in resulting error.

Examples of uncertainty:
https://stackoverflow.com/questions/22390610/the-specified-cast-from-a-materialized-system-int32-type-to-the-system-string
https://stackoverflow.com/questions/32264382/the-specified-cast-from-a-materialized-system-int32-type-to-the-system-double
https://stackoverflow.com/questions/23914429/the-specified-cast-from-a-materialized-system-datetime-type-to-the-system-str

Although the referenced issue #561 is Closed, the PR that solved that issue was not merged (because of missing tests).

This PR include changes from #771 with some additional error handling for Shaper class (specifically GetColumnValueWithErrorHandling method) also contains tests for target cases.

### DIFF
--- a/src/EntityFramework/Core/Common/EntityUtil.cs
+++ b/src/EntityFramework/Core/Common/EntityUtil.cs
@@ -327,6 +327,25 @@ namespace System.Data.Entity.Core
             }
         }
 
+        internal static InvalidOperationException ValueInvalidCast(Type valueType, Type destinationType, string columnName)
+        {
+            DebugCheck.NotNull(valueType);
+            DebugCheck.NotNull(destinationType);
+            if (destinationType.IsValueType()
+                && destinationType.IsGenericType()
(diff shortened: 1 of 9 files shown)


DenisBalan:
as @ajcvickers reviewed #771 , basically this is improved version of that PR including tests. 😉

ajcvickers:

PR #1667, opened 2020-05-11
Fix signing validation

dougbu:
- work around Arcade eng/Signing.props requirement


wtgodbe:
@dougbu we shouldn't need this after https://github.com/dotnet/ef6/pull/1668 is merged

dougbu:
Closing in favour of #1668

PR #1670, opened 2020-05-12
Update branding to 6.4.5

wtgodbe:
- Merges internal -> public
- Removes internal feeds from nuget.config
- Updates branding

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -7,11 +7,11 @@
     </Dependency>
     <Dependency Name="Microsoft.NETCore.App.Runtime.win-x64" Version="3.1.4">
       <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-core-setup</Uri>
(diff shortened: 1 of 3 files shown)


eng/Version.Details.xml
```diff
@@ -6,12 +6,12 @@
       <Sha>0f7f38c4fd323b26da10cce95f857f77f0f09b48</Sha>
     </Dependency>
     <Dependency Name="Microsoft.NETCore.App.Runtime.win-x64" Version="3.1.4">
-      <Uri>https://github.com/dotnet/core-setup</Uri>
-      <Sha>57d5bbb58f17a8cb3a82c81839c9379b4fcfe0d8</Sha>
+      <Uri>https://dev.azure.com/dnceng/internal/_git/dotnet-core-setup</Uri>
```
dougbu, line 9: I recall changing the `<Uri/>` is temporary, doesn't matter much, and will go away next time we update dependencies from core-setup. Is that correct?

wtgodbe: That's correct

PR #1706, opened 2020-07-15
Update azure-pipelines.yml

smitpatel:
Batched builds

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -19,9 +19,12 @@ variables:
       value: ''
 
 trigger:
-  - master
-  - release/*
-  - internal/release/6.*
+  batch: true
+  branches:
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -19,9 +19,12 @@ variables:
...
 trigger:
-  - master
-  - release/*
-  - internal/release/6.*
+  batch: true
+  branches:
+    include:
+      - master
+      - release/*
+      - internal/release/6.*
```
dougbu, line 27: nit: these dashes could line up with `include:`

smitpatel: Since it is subsection I would prefer to keep it indented. It also aligns with the documentation.

PR #1737, opened 2020-08-19
Feature Request : Enable MIN/MAX aggrgates on HierarchyId type

magicmoux:
Just adding overlooked HierarchyId mapping of aggregates 

Use-case : allowing for simple expression of root/leaf nodes filtering within LINQ.


ajcvickers:

PR #1770, opened 2020-11-11
Publish EF6 designer DLLs and PDBs

ryanbrandenburg:
Currently we don't Publish all the DLL's and PDB's that this build produces, which we need in order to archive them for the VS symbol-server. This should start preserving those DLL's for use in just such a manner.


ryanbrandenburg:
Closing this. We're not fixing the Symbols problem in the current PR, and none of the changes made in this PR solved the PDB production/storage anyway.

Issue #1844, opened 2021-04-07
DbModel Build performance


zbrkic:
Method [ConfigureEntityTypes](https://github.com/dotnet/ef6/blob/99c172bc465492fa37f9efa07900653975ec79b4/src/EntityFramework/ModelConfiguration/Configuration/ModelConfiguration.cs#L542) in ModelConfiguration.cs has this loop:

https://github.com/dotnet/ef6/blob/99c172bc465492fa37f9efa07900653975ec79b4/src/EntityFramework/ModelConfiguration/Configuration/ModelConfiguration.cs#L542-L547

which calls [Configure](https://github.com/dotnet/ef6/blob/99c172bc465492fa37f9efa07900653975ec79b4/src/EntityFramework/ModelConfiguration/Configuration/Types/EntityTypeConfiguration.cs#L685) for each entityType:
https://github.com/dotnet/ef6/blob/99c172bc465492fa37f9efa07900653975ec79b4/src/EntityFramework/ModelConfiguration/Configuration/Types/EntityTypeConfiguration.cs#L685-L710

[ConfigureDependentKeys](https://github.com/dotnet/ef6/blob/99c172bc465492fa37f9efa07900653975ec79b4/src/EntityFramework/ModelConfiguration/Configuration/Types/EntityTypeConfiguration.cs#L837) does not depend on entityType.

https://github.com/dotnet/ef6/blob/99c172bc465492fa37f9efa07900653975ec79b4/src/EntityFramework/ModelConfiguration/Configuration/Types/EntityTypeConfiguration.cs#L837

Can we call it only once instead of for each entityType?

When building a big DBModel, calling it only once makes a difference.

### Further technical details
EF version: 6.4.4

ajcvickers:
This issue has been closed because EF6 is no longer being actively developed. We are instead focusing on stability of the codebase, which means we will only make changes to address security issues. See the repo [README](https://github.com/dotnet/ef6/blob/main/README.md) for more information.

PR #1853, opened 2021-05-09
Configure dependent keys only once

zbrkic:


zbrkic:
Fixes #1844

zbrkic:
> Tests?

Performance tests? I cannot share my model but build time is reduced from 21 seconds to 13 seconds.

ajcvickers:

PR #2008, opened 2022-07-05
feat: add validation to prevent invalid migrations name

BirajMainali:
If we use reserve key character while adding a new migration it doesn't work as aspected for eg `dotnet ef migrations add "add: user`, this tries to create a migration file same as provided name, but we are using the reserve key characters, i don't know how but it's an unknown migrations file.

## Steps to reproduce
 - Adding initial migration
image
- Error
image

So, I've tried to fix this issue, I apologize if this brings the breaking changes.


ajcvickers:

PR #2064, opened 2022-11-26
Update to S.D.S 4.8.5

### LINKED ISSUE #2061: Update System.Data.SqlClient dependency to version 4.8.5
November 8th a vulnerability was reported to System.Data.SqlClient versions <=4.8.4 which is fixed in 4.8.5 (https://github.com/advisories/GHSA-8g2p-5pqh-5jmc).

We are using the latest EF6 (6.4.4) and we found that the it contains System.Data.SqlClient 4.8.1. No update available within the NuGet packages.

Is there a reason why this is not updated? Is EF6 not affected because of some other protections are in place? Or will there be an update soon to replace the System.Data.SqlClient?

Thanks!

ErikEJ:
fixes #2061


ajcvickers (approved):

ajcvickers:
Thanks Erik!

PR #2150, opened 2023-08-31
Merge branch 'release/6.5' => 'main'

bricelam:


ajcvickers (approved):

bricelam:
Hmm, equality between two dynamic `Assemby` instances (from System.Reflection.Emit) broke sometime between .NET 6 and .NET 8. We could work around it by adding a custom comparer to a `Dictionary<Assembly, object>` in the runtime, but I think I'll just treat it as a test bug since I don't think many customers are using dynamic assemblies. Skipping the test on .NET 8. If customers hit it, the runtime team should probably dig into why it stopped working.

PR #1883, opened 2021-07-26
Migrate license from Apache -> MIT

Pilchie:

PR #2160, opened 2023-10-13
Inform that licence is MIT instead of Apache

worming004:
It is very unlikely that we will accept a pull request for this repo. See the [README](https://github.com/dotnet/ef6/blob/main/README.md) for more info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Entity Framework Core (EF Core) is a lightweight and extensible version of Entit
 
 ## Forks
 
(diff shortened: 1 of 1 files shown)


CZEMacLeod:
I'm not sure that a **licencing**/documentation issue is a 'security' issue (as mentioned) but it is close, and, as this does not affect the codebase (which they don't want to touch), I feel like it should be taken seriously and included. Just my two pence.

bricelam (approved):
Thanks! Looks like we missed this in PR #1883

PR #2162, opened 2023-10-18
Propose switching test projects to minimum net462

mgravell:
Build-ops suggestion from discussion https://github.com/dotnet/ef6/pull/2158

Context: test runner no longer supports NET452, which is being proposed by automatic updates

This:

- changes test projects to use minimum NET462 (was NET452)
- where previous test code used `#if NET452` etc, use `#if NETFRAMEWORK` instead, to make version-independent (they are not multi-targeting multiple netfx TFMs)

Note that this PR **does not** change the TFM for the actual library, which offers down to NET45


javiercn:
@dotnet/efteam can someone take a look and sign off?

bricelam (approved):

PR #2172, opened 2023-11-09
Remove Dotnet-Blob-Feed variable from azure-pipelines.yml

akoeplinger:
See https://github.com/dotnet/arcade/issues/13963

/cc @bricelam


bricelam:
Thanks! Rebased onto release/6.5. I'll merge to main after this one goes through

PR #2209, opened 2024-02-26
remove wrong AssemblyDefaultAlias

ErikEJ:


AndriySvyryd:
Is it wrong or just unnecessary? I.e. does it cause any issues?

ErikEJ:
It is not the assembly name of the new provider, so it is "wrong" in that respect.

I am not aware of any issues, but a release of a preview could help uncover any 😉

AndriySvyryd (approved):

AndriySvyryd:
Thanks for your contibution!

PR #2263, opened 2024-06-16
fix for broken NuGet package

### LINKED ISSUE #2262: EF 6.5.0 RTM NuGet Package BROKEN
# RTM PACKAGE FAILURE

There is an error in the package for 6.5.0 that has broken the EF build process in Visual Studio. The package is not being built properly, therefore `EntityDeploy` is no longer being called to compile EDMX files into the project.

👉🏻 **For people experiencing this issue, there is a temporary workaround, see below.** 👈🏻

## SYMPTOMS:
You will suddenly get a "Metadata cannot be loaded" error in any project that leverages the `DbContext`, where the code previously functioned normally.

## IMPACT
Automated builds that were previously working are now failing. I spent several hours tracking down why my services suddenly stopped working.

## NEEDED RESOLUTION
- [ ] **EntityFramework 6.5.0 package must be deprecated on NuGet.org IMMEDIATELY**
- [ ] Revert the change from Preview 2 that broke the build folder
- [ ] The release process should be validating the package contents before deploying the package

## EVIDENCE
New EntityFramework 6.5.0 RTM vs EntityFramework 6.4.4 RTM packages:

image
image

## WORKAROUND
In the project with your EDMX file, replace your `<PackageReference Include="EntityFramework" ... />` line with the following entries:

```xml
        <PackageReference Include="Microsoft.EntityFramework.SqlServer" Version="6.*" />
        <PackageReference Include="EntityFramework" Version="6.4.*" ExcludeAssets="compile,runtime" NoWarn="NU1605" />
```
This will tell the build process to use the `EntityCompile` tasks from the EF 6.4.4 package, [but NOT to reference anything in the `lib` folder for compilation](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets). That means the downstream reference to `EntityFramework` 6.5 from the `Microsoft.EntityFramework.SqlServer` package will still get compiled into the build and your code will work correctly. The "NoWarn" attribute will tell the build process not to complain that you're referencing a version that the `.SqlServer` package doesn't like.

ErikEJ:
fixes #2262

@AndriySvyryd FYI 😭

### DIFF
--- a/src/EntityFramework.NuGet/EntityFramework.NuGet.nuspec
+++ b/src/EntityFramework.NuGet/EntityFramework.NuGet.nuspec
@@ -51,10 +51,10 @@
     <file src="../../artifacts/bin/EntityFramework.SqlServer/$Configuration$/netstandard2.1/EntityFramework.SqlServer.dll" target="lib/netstandard2.1/" />
     <file src="../../artifacts/bin/EntityFramework.SqlServer/$Configuration$/netstandard2.1/EntityFramework.SqlServer.xml" target="lib/netstandard2.1/" />
     <file src="../../artifacts/bin/Microsoft.Data.Entity.Build.Tasks/$Configuration$/netstandard2.0/Microsoft.Data.Entity.Build.Tasks.dll" target="build/" />
-    <file src="build/**/*" target="build/" />
-    <file src="buildTransitive/**/*" target="buildTransitive/" />
-    <file src="content/**/*" target="content/" />
-    <file src="tools/**/*" target="tools/" />
+    <file src="build\**\*" target="build/" />
+    <file src="buildTransitive\**\*" target="buildTransitive/" />
+    <file src="content\**\*" target="content/" />
+    <file src="tools\**\*" target="tools/" />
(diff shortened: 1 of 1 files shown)


AndriySvyryd (approved):

ErikEJ:
I have verified/reproduced locally that this change broke the package

AndriySvyryd:
@ErikEJ Please rebase on `release/6.5`

ErikEJ:
@AndriySvyryd How do I do that from my branch? (Excuse my ignorance) - with git cli

AndriySvyryd:
Something like `git rebase upstream/release/6.5 -i`
Assuming that the remote for this repo is called `upstream`

ErikEJ:
@AndriySvyryd That di not go so well - let me start over

birbilis:
got bit too by the previous NuGet package (wasn't removing old entries from files and was also creating some configuration subfolders), had to rollback via version control and install the new one which works fine (upgrading to new package wasn't solving the mess). Maybe rolling back to previous NuGet package might have worked too, but not sure

PR #2301, opened 2024-10-28
[release/6.5] Update binskim and prefast

AndriySvyryd:
Remove CodeQL references from the public definition


AndriySvyryd:
> Do we understand why other checks didn't run?

Yes, there was an AzDO outage

PR #2345, opened 2025-12-29
Add fix for .NET 10 first-class span support

ericsampson:
See:
- https://github.com/dotnet/ef6/issues/2322
- https://github.com/dotnet/runtime/issues/109757

This PR is as direct of a port of @roji's [original .EF Core fix](https://github.com/dotnet/efcore/pull/35339) as I could make it, with minor adjustments as required by EF6.

I added unit tests that demonstrated failure before the fix and success on all cases afterwards.

Thanks for the help hopefully getting this into EF6!
Cheers,
Eric

### DIFF
--- a/src/EntityFramework/Core/Objects/ELinq/LinqExpressionNormalizer.cs
+++ b/src/EntityFramework/Core/Objects/ELinq/LinqExpressionNormalizer.cs
@@ -7,6 +7,7 @@ namespace System.Data.Entity.Core.Objects.ELinq
     using System.Data.Entity.Utilities;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
     using System.Linq.Expressions;
     using System.Reflection;
 
@@ -124,6 +125,32 @@ namespace System.Data.Entity.Core.Objects.ELinq
                    ((ConstantExpression)expression).Value.Equals(0);
         }
 
+#if NETSTANDARD2_1
+        // <summary>
+        // Attempts to unwrap an implicit Span or ReadOnlySpan cast expression to get the underlying array or collection.
+        // This is used to rewrite MemoryExtensions methods that operate on Span/ReadOnlySpan back to Enumerable methods.
+        // </summary>
+        private static bool TryUnwrapSpanImplicitCast(Expression expression, out Expression result)
+        {
+            if (expression is MethodCallExpression methodCallExpr
+                && methodCallExpr.Method.Name == "op_Implicit"
+                && methodCallExpr.Method.DeclaringType != null
+                && methodCallExpr.Method.DeclaringType.IsGenericType
+                && methodCallExpr.Arguments.Count == 1)
+            {
+                var genericTypeDefinition = methodCallExpr.Method.DeclaringType.GetGenericTypeDefinition();
+                if (genericTypeDefinition == typeof(Span<>) || genericTypeDefinition == typeof(ReadOnlySpan<>))
+                {
+                    result = methodCallExpr.Arguments[0];
+                    return true;
+                }
+            }
+
+            result = null;
+            return false;
+        }
+#endif
+
         // <summary>
         // Handles MethodCall patterns:
         // - Operator overloads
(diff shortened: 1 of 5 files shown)


src/EntityFramework/Core/Objects/ELinq/LinqExpressionNormalizer.cs
```diff
@@ -132,6 +159,51 @@ private static bool IsConstantZero(Expression expression)
         [SuppressMessage("Microsoft.Maintainability", "CA1502:AvoidExcessiveComplexity")]
         internal override Expression VisitMethodCall(MethodCallExpression m)
         {
+#if NETSTANDARD2_1
+            // .NET 10 made changes to overload resolution to prefer Span-based overloads when those exist ("first-class spans").
+            // Unfortunately, the LINQ interpreter does not support ref structs, so we rewrite e.g. MemoryExtensions.Contains to
```
roji: Just to be pedantic, I'm not sure that EF6 uses the LINQ interpreter (it might - I'm really not sure). The reason to rewrite is more that EF6 simply doesn't recognize MemoryExtensions.Contains in translation, I think. But at most this means tweaking the comment here, nothing more.

test/UnitTests/UnitTests.csproj
```diff
@@ -3,10 +3,10 @@
   <PropertyGroup>
     <RootNamespace>System.Data.Entity</RootNamespace>
     <AssemblyName>EntityFramework.UnitTests</AssemblyName>
-    <TargetFramework>$(NetFrameworkToolCurrent)</TargetFramework>
+    <TargetFrameworks>$(NetFrameworkToolCurrent);$(DefaultNetCoreTargetFramework)</TargetFrameworks>
```
roji: IIUC this means we weren't running previously these unit tests on .NET Core, only on .NET Framework; the change appears to make us run **some** tests on .NET Core, even if quite a few are excluded because of various dependencies on .NET Framework.

@AndriySvyryd can you give this a look and confirm?

roji (commented):
Thanks @ericsampson, the PR generally looks quite good.

However, the bar for any changes to EF6 is very high, and as very few users have been complaining about this issue, we weren't planning on fixing this problem at this time. I'll bring this up internally and we'll see what to do.

test/UnitTests/UnitTests.csproj
ericsampson: Correct, it seemed like it "couldn't hurt" to add the extra safety of running the many tests that do work correctly in the current test code under .NET Core.

That's why I wanted to ask y'all about your preference for the unit test strategy and csproj. I could definitely tweak it so that only the single `LinqExpressionNormalizerTests.cs` file is also run under .NET Core. Just wanted to get your opinion/preferences.

ericsampson:
> the bar for any changes to EF6 is very high, and as very few users have been complaining about this issue, we weren't planning on fixing this problem at this time. I'll bring this up internally and we'll see what to do.

@roji I totally understand. I was hoping to just reduce the "activation energy" by providing a working PR for you folks to consider rather than expecting someone from MSFT to do it, as it would likely not meet prioritization or at least not anytime soon.

The build issue should be addressed. Last time the build queue was taking so long to pick up the job that I had to just walk away and let it get picked up overnight, so that's why I didn't fix the failure immediately. Looks like the fix is now in the same queue waiting game state :)

Cheers

ericsampson:
@roji CI tests are passing now :)

src/EntityFramework/Core/Objects/ELinq/LinqExpressionNormalizer.cs
```diff
@@ -132,6 +159,51 @@ private static bool IsConstantZero(Expression expression)
...
+                {
+                    // Note that MemoryExtensions.Contains has an optional 3rd ComparisonType parameter; we only match when
+                    // it's null.
+                    case nameof(MemoryExtensions.Contains)
+                        when m.Arguments.Count >= 2
+                        && (m.Arguments.Count == 2
+                            || m.Arguments.Count == 3 && m.Arguments[2] is ConstantExpression constantExpr && constantExpr.Value == null)
+                        && TryUnwrapSpanImplicitCast(m.Arguments[0], out var unwrappedSpanArg):
+                    {
+                        var containsMethod = typeof(System.Linq.Enumerable).GetMethods()
```
roji: This can be cached on a nullable private field so we don't have to perform the lookup every time (but only the first time it's needed).

roji (changes requested):
Thanks, this looks almost ready (good work!). A couple of small fixes and I'll merge it.

Just to clearly set expectations; for now, although we'll be merging this PR, we won't be releasing a new version of EF6 in the immediate future. In principle, new versions of EF6 are only being released for security fixes, which this isn't. Based on feedback we can reevaluate.

src/EntityFramework/Core/Objects/ELinq/LinqExpressionNormalizer.cs
ericsampson: Implemented using `Lazy`.

roji (approved):
Thanks @ericsampson, LGTM.

test/UnitTests/UnitTests.csproj
```diff
@@ -16,6 +16,17 @@
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
 
+  <!--
+    For .NET 10+ builds, only include Span-related tests. The vast majority of the UnitTests project
+    depends on .NET Framework-specific infrastructure (SqlServerCe, Remoting, BinaryFormatter, etc.)
+    and was not designed for cross-platform use. Rather than exclude hundreds of tests, we whitelist
```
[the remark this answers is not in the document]
roji: Can you please just tweak this word? I tried but it seems you haven't given permission to push to your PR branch.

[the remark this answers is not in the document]
ericsampson: Yeah I agree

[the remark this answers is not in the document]
ericsampson: Done!

I thought I had set things up to allow you to push changes to the fork, but I've never been able to successfully figure out how to get that to work.

ericsampson:
@roji should I merge this in? Or do you want to. What do you think of the idea of publishing this to NuGet as `6.6.1-alpha` ? It would allow people that are affected to pull down a working fix for their apps, without having to wait for an unknown amount of time before the next security patch release. Cheers

bkzoller:
This is blocking us from upgrading to C# 14. Luckily we were alerted to this problem with hundreds of test failures in the CI pipeline from simply switching from C# 13 to C# 14. This is a potential landmine for any projects using EF6 and .NET 10 together without such test coverage.

PaulVrugt:
@roji @ericsampson not sure who to tag here, but this merge is useless without a new release on nuget. What are the plans for releasing this?

ericsampson:
@PaulVrugt looks like this made it into a [6.5.2 patch release](https://github.com/dotnet/ef6/releases/tag/v6.5.2), and [NuGet](https://www.nuget.org/packages/EntityFramework/6.5.2)

thanks @roji and @AndriySvyryd! I threw in a docs PR to update the EF6 release notes there, to help people find it as well.

PaulVrugt:
> @PaulVrugt looks like this made it into a [6.5.2 patch release](https://github.com/dotnet/ef6/releases/tag/v6.5.2), and [NuGet](https://www.nuget.org/packages/EntityFramework/6.5.2)
> 
> thanks @roji and @AndriySvyryd! I threw in a docs PR to update the EF6 release notes there, to help people find it as well.

Awesome! Thanks for the heads up!

PR #2413, opened 2026-08-05
Remove obsolete MSRC storage variable group

missymessa:
Removes the stale DotNet-MSRC-Storage import from the EF6 pipeline.

VG25 now contains only the non-secret DummySecret placeholder, and DummySecret is not referenced by EF6. The latest successful internal/release build also ran without a resolved storage-feed key, so removing the group does not change the observed pipeline behavior.

Tracking: https://dev.azure.com/dnceng/internal/_workitems/edit/10812


missymessa:
@wtgodbe can you please take a look! Thanks!

wtgodbe (approved):

